PR #7, opened 2020-12-09
Env stuff, docker-compose log mount

rosalieper:


addshore (changes requested):
I'm going to remove the composer.lock change as it seemingly updates a bunch of things that we may not want to

PR #16, opened 2021-03-23
Migrate wiki_settings value column from string to text

jakobw:


addshore:
Tested the migration locally and everything seems to continue working well.
So now values here have a `64 Kilobytes` limit

PR #23, opened 2021-03-24
Add Fed Props and Manifest setting options

addshore:


app/Rules/SettingWikibaseManifestEquivEntities.php
```diff
@@ -0,0 +1,45 @@
...
+     * Determine if the validation rule passes.
+     *
+     * @param  string  $attribute
+     * @param  mixed  $value
+     * @return bool
+     */
+    public function passes($attribute, $value)
+    {
+        $value = json_decode( $value, true );
+        if($value===null) {
```
toban, line 20: what happens if its just an object without anything inside would pass?

toban: nevermind tried it locally does not seem to work.

PR #38, opened 2021-03-26
Add wgLogo to public settings

jakobw:


app/Wiki.php
```diff
@@ -79,8 +79,9 @@ public function settings()
     public function publicSettings() {
         return $this->settings()->whereIn('name',
         [
-            // FIXME: this list is evil and should be kept in sync with WikiSettingControler?!
+            // FIXME: this list is evil and should be kept in sync with WikiSettingController?!
             'wgDefaultSkin',
+            'wgLogo',
```
addshore, line 84: Aaah, this was missed as it is not actually in WikiSettingController as it has its own API.!

PR #93, opened 2021-05-14
Don't require GITHUB_TOKEN in docker related flows

addshore:


addshore:
This (and all repos) would need a different docker lint step to be able to pass

PR #122, opened 2021-07-05
Added job to create elastic indexes when a new wiki is created

bereket-WMDE:
[Bug: T285541](https://phabricator.wikimedia.org/T285541)

### DIFF
--- a/app/Http/Controllers/WikiController.php
+++ b/app/Http/Controllers/WikiController.php
@@ -6,6 +6,7 @@ use App\Jobs\KubernetesIngressCreate;
 use App\Jobs\MediawikiInit;
 use App\Jobs\ProvisionQueryserviceNamespaceJob;
 use App\Jobs\ProvisionWikiDbJob;
+use App\Jobs\ElasticSearchIndexInit;
 use App\QueryserviceNamespace;
 use App\Wiki;
 use App\WikiDb;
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/WikiController.php
```diff
@@ -108,6 +116,8 @@ public function create(Request $request): \Illuminate\Http\Response
         $this->dispatch(new ProvisionWikiDbJob(null, null, 10));
         $this->dispatch(new ProvisionQueryserviceNamespaceJob(null, 10));
 
+        $this->dispatch(new ElasticSearchIndexInit($wiki->domain));
```
bereket-WMDE, line 119: @toban 
Do we need to create this class somewhere ?

bereket-WMDE: i guess yes. since psalm is failing the lint test

toban: Looks like we might just need to import it!

toban (approved):

app/Http/Controllers/WikiController.php
```diff
@@ -79,6 +80,14 @@ public function create(Request $request): \Illuminate\Http\Response
                 'value' => Str::random(64),
             ]);
 
+            // Enable elasticsearch for new wikis by default
+            // T285541
+            WikiSetting::create([
```
addshore, line 85: For now I don't think we want to do this!
We should leave the default to be false.
We do however still want to be able to call the API that we made, when the thing is disabled.

addshore (changes requested):

addshore (approved):

PR #147, opened 2021-08-10
[ELASTICSEARCH-FIRST] Enable Elasticsearch for new wikis

toban:
This enables elasticsearch for new wikis through the ElasticSearchIndexInit job.

Lexemes and Items end up in the same index so no special treatment should be needed when this setting is enabled but probably needs https://github.com/wbstack/mediawiki/pull/121 to be deployed before.

### DIFF
--- a/app/Http/Controllers/WikiController.php
+++ b/app/Http/Controllers/WikiController.php
@@ -80,7 +80,7 @@ class WikiController extends Controller
                 'value' => Str::random(64),
             ]);
 
-            // Create elasticsearch setting to be updated once ElasticSearch has been initialized
+            // Create the enabled elasticsearch setting
             // T285541
             WikiSetting::create([
                 'wiki_id' => $wiki->id,
--- a/app/Http/Controllers/WikiController.php
+++ b/app/Http/Controllers/WikiController.php
@@ -118,7 +118,7 @@ class WikiController extends Controller
         $this->dispatch(new ProvisionQueryserviceNamespaceJob(null, 10));
 
         // dispatch elasticsearch init job to enable the feature
-        $this->dispatch(new ElasticSearchIndexInit($wiki->domain, $wiki->id));
+        $this->dispatch(new ElasticSearchIndexInit($wiki->id));
 
         $res['success'] = true;
         $res['message'] = 'Success!';
--- a/app/Http/Controllers/WikiController.php
+++ b/app/Http/Controllers/WikiController.php
@@ -80,13 +80,13 @@ class WikiController extends Controller
                 'value' => Str::random(64),
             ]);
 
-            // Enable elasticsearch for new wikis by default
+            // Create elasticsearch setting to be updated once ElasticSearch has been initialized
             // T285541
-//             WikiSetting::create([
-//                 'wiki_id' => $wiki->id,
-//                 'name' => 'wwExtEnableElasticSearch',
-//                 'value' => true,
-//             ]);
+            WikiSetting::create([
+                'wiki_id' => $wiki->id,
+                'name' => 'wwExtEnableElasticSearch',
+                'value' => false,
+            ]);
 
             // Also track the domain forever in your domains table
             $wikiDomain = WikiDomain::create([
@@ -117,7 +117,8 @@ class WikiController extends Controller
         $this->dispatch(new ProvisionWikiDbJob(null, null, 10));
         $this->dispatch(new ProvisionQueryserviceNamespaceJob(null, 10));
 
-//        $this->dispatch(new ElasticSearchIndexInit($wiki->domain));
+        // dispatch elasticsearch init job to enable the feature
+        $this->dispatch(new ElasticSearchIndexInit($wiki->domain, $wiki->id));
 
         $res['success'] = true;
         $res['message'] = 'Success!';
--- a/app/Http/Controllers/WikiController.php
+++ b/app/Http/Controllers/WikiController.php
@@ -76,7 +76,7 @@ class WikiController extends Controller
             // Docs: https://www.mediawiki.org/wiki/Manual:$wgSecretKey
             WikiSetting::create([
                 'wiki_id' => $wiki->id,
-                'name' => 'wgSecretKey',
+                'name' => WikiSetting::wgSecretKey,
                 'value' => Str::random(64),
             ]);
 
@@ -84,8 +84,8 @@ class WikiController extends Controller
             // T285541
             WikiSetting::create([
                 'wiki_id' => $wiki->id,
-                'name' => 'wwExtEnableElasticSearch',
-                'value' => false,
+                'name' => WikiSetting::wwExtEnableElasticSearch,
+                'value' => true,
             ]);
 
             // Also track the domain forever in your domains table
--- a/app/Jobs/ElasticSearchIndexInit.php
+++ b/app/Jobs/ElasticSearchIndexInit.php
@@ -6,20 +6,18 @@ use App\WikiSetting;
 use App\Http\Curl\CurlRequest;
 use App\Http\Curl\HttpRequest;
 use Illuminate\Support\Facades\Log;
+use App\Wiki;
 
 class ElasticSearchIndexInit extends Job implements ShouldBeUnique
 {
-    private $wikiDomain;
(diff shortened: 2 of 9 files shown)


toban:
This still isn't working in the locally in the mediawiki repo. We are probably missing some configuration that needs to be added.

toban:
Should have started in the other end on this epic, seems elasticsearch does not need any special updating to handle lexeme.

https://github.com/wbstack/mediawiki/pull/121

toban:
this is being re-written not to use the observer

app/Providers/EventServiceProvider.php
```diff
@@ -27,7 +29,5 @@ class EventServiceProvider extends ServiceProvider
     public function boot()
     {
         parent::boot();
-
-        //
```
toban, line 31: fix

tests/Jobs/ElasticSearchIndexInitTest.php
```diff
@@ -0,0 +1,105 @@
...
+        WikiSetting::factory()->create(
+            [
+                'wiki_id' => $this->wiki->id,
+                'name' => WikiSetting::wwExtEnableElasticSearch,
+                'value' => false
+            ]
+        );
+    }
+
+    public function testSuccess()
```
toban, line 36: Add test for when no settings

tests/Routes/Wiki/CreateTest.php
```diff
@@ -4,10 +4,62 @@
...
     protected $route = 'wiki/create';
 
     use OptionsRequestAllowed;
+    use DatabaseTransactions;
+
+    public function testWikiCreateDispatchesSomeJobs()
+    {
+        Queue::fake();
+
+        // This random user probably doesn't exist in the db
```
toban, line 26: remove

app/Jobs/ElasticSearchIndexInit.php
```diff
@@ -63,15 +84,40 @@ public function handle()
...
+        if ( in_array( $newlyCreated,  $output ) ) {
+
+            // newly created index succeeded, turn on the wiki setting
+            $enableElasticSearchFeature = true;
+
+        } else if ( in_array( $updated, $output ) ) {
+
+            // script ran and update was successful, make sure feature is enabled
+            $enableElasticSearchFeature = true;
         }
```
lucaswerkmeister, line 114: We probably want to log some kind of warning if neither of these conditions is true?

app/Http/Controllers/WikiController.php
```diff
@@ -80,13 +80,13 @@ public function create(Request $request): \Illuminate\Http\Response
...
-//             WikiSetting::create([
-//                 'wiki_id' => $wiki->id,
-//                 'name' => 'wwExtEnableElasticSearch',
-//                 'value' => true,
-//             ]);
+            WikiSetting::create([
+                'wiki_id' => $wiki->id,
+                'name' => 'wwExtEnableElasticSearch',
+                'value' => false,
+            ]);
```
lucaswerkmeister, line 89: If we wanted to ensure that no pages get created before ElasticSearch is ready (because we don’t want any content to bypass the index), we could set some internal setting that is then assigned to [`$wgReadOnly`](https://www.mediawiki.org/wiki/Manual:$wgReadOnly), so that the wiki only becomes writable once the search index is ready (i.e. change that setting in the init job as well). But right now we probably don’t need that.

lucaswerkmeister (approved):

app/Jobs/ElasticSearchIndexInit.php
toban: done

lucaswerkmeister (approved):

app/Http/Controllers/WikiController.php
addshore: Indeed, and this may be something that we want to think about.
We could also force indexing of the whole site after enabling wwExtEnableElasticSearch / running the job to create the indexes?

app/Jobs/ElasticSearchIndexInit.php
```diff
@@ -1,49 +1,82 @@
...
+            $this->fail(
+                new \RuntimeException('wbstackElasticSearchInit call for '.$this->wikiDomain.' was triggered but not setting available.')
+            );
+
+            return;
+        }
+        
+        $this->request->setOptions( 
+            [
+                CURLOPT_URL => getenv('PLATFORM_MW_BACKEND_HOST').'/w/api.php?action=wbstackElasticSearchInit&format=json',
```
addshore, line 55: If the setting is false, then won't this API call fail? as elasticsearch will not be enabled?
Or am I missing something?

toban: Do you think it would be ok to enable it on creation and have the job turn it off it it fails?

This would leave the site without working search until the job has either completed or failed.

toban: I guess we could enable / disable it in the job too but not sure if it would be necessary

app/Http/Controllers/WikiController.php
toban: In the current case it wouldn't make much sense since the site will probably be pretty empty?

When this setting would be enabled for something that already has a lot of data it would make more sense to me to run the ForceSearchIndex thing

app/Jobs/ElasticSearchIndexInit.php
addshore: > Do you think it would be ok to enable it on creation and have the job turn it off if it fails?
> 
> This would leave the site without working search until the job has either completed or failed.

I think this should be fine for now.

app/Jobs/ElasticSearchIndexInit.php
```diff
@@ -52,15 +85,46 @@ public function handle()
...
             );
 
-            return; //safegaurd
+            return;
         }
+
+        $output = $response['wbstackElasticSearchInit']['output'];
+
+        $newlyCreated = "\tCreating index...ok"; // occurs a couple of times when newly created
+        $updated = "\t\tValidating {$this->wikiDomain}_general alias...ok"; // occurs on a successful update run
```
addshore, line 102: This should probably be the wiki db name, or the wiki id.

tests/Jobs/ElasticSearchIndexInitTest.php
```diff
@@ -0,0 +1,115 @@
...
+        parent::setUp();
+
+        $this->user = User::factory()->create(['verified' => true]);
+        $this->wiki = Wiki::factory()->create();
+        WikiManager::factory()->create(['wiki_id' => $this->wiki->id, 'user_id' => $this->user->id]);
+        WikiSetting::factory()->create(
+            [
+                'wiki_id' => $this->wiki->id,
+                'name' => WikiSetting::wwExtEnableElasticSearch,
+                'value' => false
```
addshore, line 31: This could maybe se set to true, though that would make the assertion further down harder.
The assertion would need to switch to checking if the job was successful instead?

OR keep this as false, but perhaps leave a note about why? (So someone else doesn't set it to true and essentially eliminate the assertion of this test.

addshore (changes requested):

app/Jobs/ElasticSearchIndexInit.php
```diff
@@ -1,49 +1,82 @@
...
+
+            return;
+        }
+        
+        $this->request->setOptions( 
+            [
+                CURLOPT_URL => getenv('PLATFORM_MW_BACKEND_HOST').'/w/api.php?action=wbstackElasticSearchInit&format=json',
+                CURLOPT_RETURNTRANSFER => true,
+                CURLOPT_ENCODING => '',
+                CURLOPT_TIMEOUT => 10,
```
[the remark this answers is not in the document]
toban: done

tests/Jobs/ElasticSearchIndexInitTest.php
toban: done

app/Jobs/ElasticSearchIndexInit.php
toban: done

addshore (approved):
Looking good

PR #152, opened 2021-08-16
Elasticsearch deletion

toban:


toban:
Could add integration style test here like with wdqs service

toban:
> Could add integration style test here like with wdqs service

done!

.github/workflows/composer.test.yml
```diff
@@ -44,3 +44,6 @@ jobs:
       run: docker-compose exec -T api vendor/bin/phpunit
     - name: Psalm
       run: docker-compose exec -T api vendor/bin/psalm
+
+    - name: Run elasticsearch index deletion integration test
+      run: docker-compose exec -e RUN_PHPUNIT_INTEGRATION_TEST=1 -e ELASTICSEARCH_HOST=elasticsearch.svc:9200 -T api vendor/bin/phpunit tests/Jobs/Integration/ElasticSearchIndexDeleteTest.php
```
addshore, line 49: Similar to https://github.com/wbstack/api/pull/155#discussion_r694860962 it might be nice to be able to run multiple integrationy tests here rather than listing them one by one!

(But we can do that later.)

app/Jobs/ElasticSearchIndexDelete.php
```diff
@@ -0,0 +1,159 @@
...
+        }
+        $elasticSearchBaseName = $this->wikiDomain;
+        $elasticSearchHost = getenv('ELASTICSEARCH_HOST');
+        
+        if( !$elasticSearchHost ) {
+            $this->fail( new \RuntimeException('ELASTICSEARCH_HOST not configured') );
+            return;
+        }
+
+        $url = $elasticSearchHost."/_cat/indices/{$elasticSearchBaseName}*?v&s=index&h=index";
```
addshore, line 59: So this is probably fine for now, but introduces another connection between the API and an a service that is only really intended for mediawiki at this point (elasticsearch).

In terms of https://wbstack.github.io/docs/assets/img/2021-tech-overview-gce-k8s.drawio.50fd9b06.svg this would introduce another arrow / cross dependency.
The only other major one of these right now is the creation of wiki dbs in the first place, and at some point we would want to push all of that to the MediaWiki side of things, and the API would just talk to mediawiki.

app/Jobs/ElasticSearchIndexDelete.php
```diff
@@ -0,0 +1,159 @@
...
+        $url = $elasticSearchHost."/{$elasticSearchBaseName}*";
+
+        $this->request->reset();
+
+        $this->request->setOptions( 
+            [
+                CURLOPT_URL => $url,
+                CURLOPT_RETURNTRANSFER => true,
+                CURLOPT_ENCODING => '',
+                CURLOPT_TIMEOUT => 10,
```
addshore, line 129: Probably needs to be 60 seconds perhaps? (If this is blocking)

app/Jobs/ElasticSearchIndexDelete.php
```diff
@@ -0,0 +1,159 @@
...
+            );
+
+            return;
+        }
+
+        $response = json_decode($rawResponse, true);
+
+        if (!array_key_exists('acknowledged', $response) || $response['acknowledged'] !== true) {
+            $this->fail(
+                new \RuntimeException('wbstackElasticSearchDelete call for '.$this->wikiDomain.' was not successful:'.$rawResponse)
```
addshore, line 151: `wbstackElasticSearchDelete` is confusing here, as to me this says that there is an API that is being called, but this now calls elasticsearch directly.

tests/Jobs/ElasticSearchIndexDeleteTest.php
```diff
@@ -0,0 +1,79 @@
...
+{
+    use DatabaseTransactions;
+
+    private $wiki;
+    private $user;
+
+    public function setUp(): void {
+        parent::setUp();
+
+        $this->user = User::factory()->create(['verified' => true]);
```
addshore, line 24: If we start using the dbname, these tests will also have to create a wikiDB entry.

addshore (changes requested):
On the whole looks good, but ideally we would want to push the index deletion all the way down to mediawiki or elasticsearch at some point

- In the cirrus extensions as code and a maintenance script
- In wbstack mediawiki in an API that is called

But for now we can probably just leave it here and fix the other things
(especially the domain -> dbname thing)

app/Jobs/ElasticSearchIndexDelete.php
toban: done

app/Jobs/ElasticSearchIndexDelete.php
toban: done

tests/Jobs/ElasticSearchIndexDeleteTest.php
toban: done

app/Jobs/ElasticSearchIndexDelete.php
```diff
@@ -0,0 +1,157 @@
...
+class ElasticSearchIndexDelete extends Job implements ShouldBeUnique
+{
+    private $wikiId;
+
+    private $request;
+
+    /**
+     * @return void
+     */
+    public function __construct( int $wikiId, HttpRequest $request = null )
```
addshore, line 19: Will have the same HttpRequest $request injection issue

addshore (changes requested):

app/Jobs/ElasticSearchIndexDelete.php
toban: done

addshore (approved):

PR #154, opened 2021-08-17
[FIRST] Add CurlRequest + HttpRequest inteface

toban:
Abstract curl requests to an interface for DI


addshore:
I'm on the fence around the namespacing of these clases being in `http`, as they touch on a slightly different topic.

And we now have calls from the `App/Job` namespace into `App/Http`, though only into the `Curl` package, so maybe this is fine?

PR #155, opened 2021-08-17
[STANDALONE] Add job for deleting blazegraph ns

toban:
- adds test for provisioning too


app/Jobs/DeleteQueryserviceNamespaceJob.php
```diff
@@ -0,0 +1,76 @@
+<?php
+
+namespace App\Jobs;
+
+use App\QueryserviceNamespace;
+use App\Http\Curl\CurlRequest;
+use App\Http\Curl\HttpRequest;
+
+class DeleteQueryserviceNamespaceJob extends Job
```
toban, line 9: implements ShouldBeUnique

app/Jobs/DeleteQueryserviceNamespaceJob.php
```diff
@@ -0,0 +1,76 @@
...
+        $queryServiceHost = config('app.queryservice_host');
+
+        $this->request->setOptions([
+            CURLOPT_URL => $queryServiceHost.'/bigdata/namespace/' . $qsNamespace->namespace ,
+            CURLOPT_RETURNTRANSFER => true,
+            CURLOPT_ENCODING => '',
+            CURLOPT_TIMEOUT => 10,
+            CURLOPT_HTTP_VERSION => CURL_HTTP_VERSION_1_1,
+            // User agent is needed by the query service...
+            CURLOPT_USERAGENT => 'WBStack ProvisionQueryserviceNamespaceJob',
```
toban, line 44: this is wrong probably

toban: changed to DeleteQueryserviceNamespaceJob, not sure if there is any checks for this?

app/Jobs/DeleteQueryserviceNamespaceJob.php
```diff
@@ -0,0 +1,95 @@
...
+     * @return void
+     */
+    public function handle()
+    {
+        $wiki = Wiki::withTrashed()->where('id', $this->wikiId)->first();
+
+        if( !$wiki ) {
+            $this->fail( new \RuntimeException("Wiki {$this->wikiId} not found.") );
+            return;
+        }
```
lucaswerkmeister, line 45: The `$wiki` doesn’t seem to be used for anything, as far as I can tell… is the error that’s being thrown important? I assume if the wiki doesn’t exist, we would still get an error a few lines below when `$qsNamespace` can’t be found.

app/Jobs/ProvisionQueryserviceNamespaceJob.php
```diff
@@ -75,31 +81,33 @@ public function handle()
...
             // TODO error if $qsns is not actually created...
+            } else if( $response === 'EXISTS: ' .$this->namespace ) {
+                $qsns = QueryserviceNamespace::updateOrCreate([
+                    'namespace' => $this->namespace,
+                    //'internalHost' => $this->internalHost, TODO is this required?
+                    'backend' => $queryServiceHost,
+                ]);
             } else {
                 $this->fail(
                     new \RuntimeException('Valid response, but couldn\'t find "CREATED: " in: '.$response)
```
lucaswerkmeister, line 107: maybe change to CREATED or EXISTS?

app/Jobs/DeleteQueryserviceNamespaceJob.php
toban: yep true, lets remove it

app/Jobs/ProvisionQueryserviceNamespaceJob.php
toban: done

lucaswerkmeister (approved):

app/Jobs/DeleteQueryserviceNamespaceJob.php
addshore: There are not checks for this, but the UAs do show up in logs, which makes a useful name good!

.github/workflows/composer.test.yml
```diff
@@ -44,3 +44,6 @@ jobs:
       run: docker-compose exec -T api vendor/bin/phpunit
     - name: Psalm
       run: docker-compose exec -T api vendor/bin/psalm
+
+    - name: Run blazegraph integration test
+      run: docker-compose exec -e RUN_PHPUNIT_INTEGRATION_TEST=1 -T api vendor/bin/phpunit tests/Jobs/Integration/QueryserviceNamespaceJobTest.php
```
addshore, line 49: Make make sense to make this more generic and just pass the path of `tests/Jobs/Integration` and name this `Run integration tests` or something like that?
But maybe to do that properly the test dir would need a bit of a restructure?
Could probably leave this for later...

app/Jobs/DeleteQueryserviceNamespaceJob.php
```diff
@@ -0,0 +1,87 @@
+<?php
+
+namespace App\Jobs;
+
+use App\QueryserviceNamespace;
+use App\Http\Curl\CurlRequest;
+use App\Http\Curl\HttpRequest;
+use Illuminate\Contracts\Queue\ShouldBeUnique;
+
+class DeleteQueryserviceNamespaceJob extends Job implements ShouldBeUnique
```
addshore, line 10: At some point I think it will make sense to group these jobs a bit more, for example we have a provision and a delete for query service namespaces.
Similar for elastic search and wiki dbs etc.

app/Jobs/ProvisionQueryserviceNamespaceJob.php
```diff
@@ -75,31 +81,33 @@ public function handle()
...
-
-            return; //safegaurd
+            $this->fail( new \RuntimeException('cURL Error #:'.$err) );
+            return;
         } else {
             if ($response === 'CREATED: '.$this->namespace) {
                 $qsns = QueryserviceNamespace::create([
                     'namespace' => $this->namespace,
-                    //'internalHost' => $this->internalHost,
+                    //'internalHost' => $this->internalHost, TODO is this required?
```
addshore, line 95: Not right now, we could probably nuke this line.

app/Jobs/ProvisionQueryserviceNamespaceJob.php
```diff
@@ -75,31 +81,33 @@ public function handle()
...
                     'backend' => $queryServiceHost,
                 ]);
             // TODO error if $qsns is not actually created...
+            } else if( $response === 'EXISTS: ' .$this->namespace ) {
```
addshore, line 99: I don't think we want to have this as a "good case".
If the NS already exists then that hints that we have a conflict somehow.

As the namespace comes from a random string this could indiacte we havn't been as random as we expect.
https://github.com/wbstack/api/pull/155/files#diff-d92a008b006bf2d371a959dee8f8d3be5724b5fce1655f46102404cbe4f27977R33

If this were to happen, but we count this as acceptable, wikis could end up sharing a namespace?

tests/Jobs/Integration/QueryserviceNamespaceJobTest.php
```diff
@@ -0,0 +1,61 @@
...
+ * 
+ * Example: docker-compose exec -e RUN_PHPUNIT_INTEGRATION_TEST=1 -T api vendor/bin/phpunit tests/Jobs/Integration/QueryserviceNamespaceJobTest.php
+ */
+class QueryserviceNamespaceJobTest extends TestCase
+{
+    use DatabaseTransactions;
+
+    public function testIntegrationCreate()
+    {
+        if ( !getenv('RUN_PHPUNIT_INTEGRATION_TEST') ) {
```
addshore, line 29: Might be nicer / cleaner to have the env var check in setUp or setUpBeforeClass to keep it seperate from the actuall test case / cases if more are added

addshore (changes requested):

app/Jobs/DeleteQueryserviceNamespaceJob.php
toban: Yeah for sure, they should end up in separate namespaces when everything is merged

app/Jobs/ProvisionQueryserviceNamespaceJob.php
toban: Ok, 

What do you think we should do, fail the job and log an error?

tests/Jobs/Integration/QueryserviceNamespaceJobTest.php
toban: good idea

app/Jobs/ProvisionQueryserviceNamespaceJob.php
toban: done

app/Jobs/ProvisionQueryserviceNamespaceJob.php
toban: I made it fail and log an error now

app/Jobs/DeleteQueryserviceNamespaceJob.php
```diff
@@ -0,0 +1,87 @@
...
+
+        $this->request->close();
+
+        if ($err) {
+            $this->fail( new \RuntimeException('cURL Error #:'.$err) );
+            return;
+        } else {
+            if ($response === 'DELETED: '.$qsNamespace->namespace) {
+
+                QueryserviceNamespace::where([
```
addshore, line 73: We should just delete the `$qsNamespace` that we already have here.

app/Jobs/ProvisionQueryserviceNamespaceJob.php
addshore: Perfect!

addshore (changes requested):

app/Jobs/DeleteQueryserviceNamespaceJob.php
```diff
@@ -0,0 +1,87 @@
...
+
+class DeleteQueryserviceNamespaceJob extends Job implements ShouldBeUnique
+{
+    private $wikiId;
+    private $request;
+
+    /**
+     * @return void
+     */
+    public function __construct( $wikiId, HttpRequest $request = null)
```
addshore, line 18: HttpRequest can't be injected here (as discovered today)

app/Jobs/DeleteQueryserviceNamespaceJob.php
```diff
@@ -0,0 +1,87 @@
...
+            return;
+        }
+
+        $queryServiceHost = config('app.queryservice_host');
+
+        $this->request->setOptions([
+            CURLOPT_URL => $queryServiceHost.'/bigdata/namespace/' . $qsNamespace->namespace ,
+            CURLOPT_RETURNTRANSFER => true,
+            CURLOPT_ENCODING => '',
+            CURLOPT_TIMEOUT => 10,
```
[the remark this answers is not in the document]
toban: 60 seconds now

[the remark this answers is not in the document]
toban: adding this instead CURLOPT_TIMEOUT_DELETE_QUERYSERVICE_NAMESPACE and defaults to 100

app/Jobs/DeleteQueryserviceNamespaceJob.php
```diff
@@ -0,0 +1,87 @@
...
+
+        if( !$qsNamespace ) {
+            $this->fail( new \RuntimeException("Namespace for wiki {$this->wikiId} not found.") );
+            return;
+        }
+
+        $queryServiceHost = config('app.queryservice_host');
+
+        $this->request->setOptions([
+            CURLOPT_URL => $queryServiceHost.'/bigdata/namespace/' . $qsNamespace->namespace ,
```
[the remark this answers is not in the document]
toban: done

app/Jobs/DeleteQueryserviceNamespaceJob.php
toban: done

app/Jobs/DeleteQueryserviceNamespaceJob.php
toban: done

app/Jobs/ProvisionQueryserviceNamespaceJob.php
```diff
@@ -12,12 +15,15 @@ class ProvisionQueryserviceNamespaceJob extends Job
 {
     private $namespace;
     private $maxFree;
+    private $request;
 
     /**
      * @return void
      */
-    public function __construct($namespace = null, $maxFree = null)
+    public function __construct($namespace = null, $maxFree = null, HttpRequest $request = null)
```
addshore, line 23: HttpRequest cant be injected in constructor

toban: done

app/Jobs/DeleteQueryserviceNamespaceJob.php
```diff
@@ -0,0 +1,79 @@
...
+    {
+        return strval($this->wikiId);
+    }
+
+    /**
+     * @return void
+     */
+    public function handle( HttpRequest $request )
+    {
+        $qsNamespace = QueryserviceNamespace::whereWikiId($this->wikiId)->first();
```
addshore, line 37: We could / should check the delete flag here as in the elastic delete job

```
$wiki = Wiki::withTrashed()->where( [ 'id' => $this->wikiId ] )->with('settings')->with('wikiDb')->first();
```

addshore:
LGTM other than this last comment

app/Jobs/DeleteQueryserviceNamespaceJob.php
toban: done

PR #156, opened 2021-08-17
[SECOND] Add integration setup

toban:
Adds a docker-compose.integration.yml file for the api setup that can be run in CI

### DIFF
--- a/docker-compose.integration.yml
+++ b/docker-compose.integration.yml
@@ -1,14 +1,14 @@
 version: '2'
 services:
   wdqs:
-    image: wikibase/wdqs:0.3.40
+    image: ghcr.io/wbstack/queryservice:0.3.6_0.6
     volumes:
       - query-service-data:/wdqs/data
     environment:
       - WIKIBASE_HOST=default.web.mw.localhost:8080
       - WDQS_HOST=wdqs.mw.localhost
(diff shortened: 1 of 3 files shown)


docker-compose.integration.yml
```diff
@@ -0,0 +1,19 @@
+version: '2'
+services:
+  wdqs:
+    image: wikibase/wdqs:0.3.40
```
addshore, line 4: I wonder if this should infact use the wbstack build image from the query service repo?

docker-compose.integration.yml
```diff
@@ -0,0 +1,19 @@
...
+    volumes:
+      - query-service-data:/wdqs/data
+    environment:
+      - WIKIBASE_HOST=default.web.mw.localhost:8080
+      - WDQS_HOST=wdqs.mw.localhost
+      - WDQS_PORT=9999
+      - WDQS_ENTITY_NAMESPACES=120,122
```
addshore, line 11: Also lexeme namespace in this setup I guess (when lexeme is enabled)

addshore (changes requested):

addshore:
Creating a merge commit, instead of squashing here, due to all the depending on things etc.
It should make github behave nicer

PR #157, opened 2021-08-17
[STANDALONE] Delete wiki db

toban:


app/Jobs/DeleteWikiDbJob.php
```diff
@@ -0,0 +1,55 @@
...
+
+namespace App\Jobs;
+
+use App\WikiDb;
+use Illuminate\Support\Facades\DB;
+
+/**
+ * Deletes the created Database, User and WikiDB relation for a wiki 
+ */
+class DeleteWikiDbJob extends Job
```
toban, line 11: implements ShouldBeUnique probably

toban: and change id property to id

app/Jobs/DeleteWikiDbJob.php
```diff
@@ -0,0 +1,66 @@
...
+        $conn = DB::connection('mw');
+        if (! $conn instanceof \Illuminate\Database\Connection) {
+            $this->fail(new \RuntimeException('Must be run on a PDO based DB connection'));
+            return;
+        }
+        $pdo = $conn->getPdo();
+
+        // DELETE the database
+        if ($pdo->exec('DROP DATABASE '.$wikiDB->name) === false) {
+            $this->fail( new \RuntimeException('Failed to create database with dbname: '.$wikiDB->name) );
```
lucaswerkmeister, line 55:
```suggestion
            $this->fail( new \RuntimeException('Failed to drop database with dbname: '.$wikiDB->name) );
```

app/Jobs/DeleteWikiDbJob.php
```diff
@@ -0,0 +1,66 @@
...
+            // let it pass through and try to delete the user too
+        }
+
+        if ($pdo->exec('DROP USER '.$wikiDB->user) === false) {
+            $this->fail( new \RuntimeException('Failed to delete database with dbname: '.$wikiDB->name) );
```
lucaswerkmeister, line 60:
```suggestion
            $this->fail( new \RuntimeException('Failed to delete user for dbname: '.$wikiDB->name) );
```

(not sure if logging `$wikiDB->name` or `$wikiDB-user` makes more sense)

lucaswerkmeister (approved):

addshore:
In all honesty it scares me that this Job for DROPing is fairly easily called.

Perhaps we should do this in 2 steps, as this is the source of truth, just in case

1) Rename the DB and user to deleted_mwdb_* and deleted_mwuser_* (or whatever looks right)
2) Have something separate and not so automated to clean these up, that could be also automated, just with some gap, or just manually run?

addshore (changes requested):
Yes, lets remove the DROP, and instead rename the table here, and leave the final final final clean-up for some yet to be defined process..

toban:
> Yes, lets remove the DROP, and instead rename the table here, and leave the final final final clean-up for some yet to be defined process..

Changed it to renaming things instead now.

app/Jobs/DeleteWikiDbJob.php
```diff
@@ -0,0 +1,97 @@
...
+
+        $pdo = $conn->getPdo();
+        $sm = $conn->getDoctrineSchemaManager();
+        $tableNames = $sm->listTableNames();
+
+        $deletedDatabaseName = 'deleted_' . $wikiDB->name;
+
+        try {
+            $pdo->beginTransaction();
+            $pdo->exec('SET GLOBAL FOREIGN_KEY_CHECKS=0;');
```
addshore, line 75: This might not be needed, as none of these tables should have FK constraints, but also no harm in keeping it.

app/Jobs/DeleteWikiDbJob.php
```diff
@@ -0,0 +1,97 @@
...
+
+            // Create the new database
+            $pdo->exec(sprintf('CREATE DATABASE %s', $deletedDatabaseName));
+
+            foreach($tableNames as $table) {
+                $pdo->exec(sprintf('RENAME TABLE %s.%s TO %s.%s', $wikiDB->name, $table, $deletedDatabaseName, $table));
+            }
+
+            $pdo->exec(sprintf('RENAME USER %s TO %s', $wikiDB->user, 'deleted_' . $wikiDB->user ));
```
addshore, line 84: Reflecting a bit more on the situation of the USER here, maybe there is no point in keeping the user.
Since we create a new DB here to move the tables into the user wouldn't have access to anything anyway.
And in the next job, where we delete the login credentials we wouldn't be able to use the user either.

https://github.com/wbstack/api/blob/main/app/Jobs/ProvisionWikiDbJob.php#L133

So perhaps we just delete the user outright?

As a DB root we would then still be able to restore the data if we wanted it recovered for however we decide to keep the DB around for.

app/Jobs/DeleteWikiDbJob.php
```diff
@@ -0,0 +1,97 @@
...
+
+            $pdo->exec('SET GLOBAL FOREIGN_KEY_CHECKS=1;');
+            $pdo->commit();
+        } catch (PDOException $e) {
+            $pdo->rollBack();
+            $this->fail( new \RuntimeException('Failed to soft-soft delete '.$wikiDB->name . ': ' . $e->getMessage()) );
```
addshore, line 90: soft-soft, love it

app/Jobs/DeleteWikiDbJob.php
```diff
@@ -0,0 +1,97 @@
...
+
+        try {
+            $pdo->beginTransaction();
+            $pdo->exec('SET GLOBAL FOREIGN_KEY_CHECKS=0;');
+
+            // Create the new database
+            $pdo->exec(sprintf('CREATE DATABASE %s', $deletedDatabaseName));
+
+            foreach($tableNames as $table) {
+                $pdo->exec(sprintf('RENAME TABLE %s.%s TO %s.%s', $wikiDB->name, $table, $deletedDatabaseName, $table));
```
addshore, line 81: In order to make this a little easier to work with post deletion, I wonder if we could strip the prefixes from the table names (as we are renaming them all individually anyway), as these are also recorded in the wiki_db data that we will be deleting.

app/Jobs/DeleteWikiDbJob.php
```diff
@@ -0,0 +1,97 @@
...
+        if (! $conn instanceof \Illuminate\Database\Connection) {
+            $this->fail(new \RuntimeException('Must be run on a PDO based DB connection'));
+            return;
+        }
+
+        $pdo = $conn->getPdo();
+        $sm = $conn->getDoctrineSchemaManager();
+        $tableNames = $sm->listTableNames();
+
+        $deletedDatabaseName = 'deleted_' . $wikiDB->name;
```
addshore, line 71: In order to make this a tiny bit easier to work with once we delete the wiki_db data it might be nice to change the name to be something slightly different.

Perhaps

`delete_wiki_<timestamp>_<wiki_id>`

addshore (changes requested):
Sorry for the bunch of new thoughts on this one, I think it'll be ready after the next round though!

app/Jobs/DeleteWikiDbJob.php
toban: removed this part

app/Jobs/DeleteWikiDbJob.php
toban: dropping the user now

app/Jobs/DeleteWikiDbJob.php
toban: updated it to do this instead now.

app/Jobs/DeleteWikiDbJob.php
toban: it's now stripping the prefixes from tables

addshore: <3

app/Jobs/DeleteWikiDbJob.php
```diff
@@ -0,0 +1,104 @@
...
+
+        try {
+            $pdo->beginTransaction();
+
+            // Create the new database
+            $pdo->exec(sprintf('CREATE DATABASE %s', $deletedDatabaseName));
+
+            // iterate over each table and replace the prefix
+            // and move it to the deleted database by using RENAME TABLE
+            foreach($tableNames as $table) {
```
addshore, line 83: At this point $tableNames is all tables in all databases on the `mw` connection as far as I can tell?
So in this loop we would actually expect `$replacedCount !== 1` to happen, as this will be for all of the tables that do not belong to the current wiki we are trying to delete?

Rather than doing this with only strings perhaps we could use https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/schema-manager.html#listtabledetails ?

This is probably a case that can be added to the test case

At some point looping through ALL tables on the DB might end up creating a bit of overhead we want to avoid.

addshore (changes requested):

addshore (approved):

PR #158, opened 2021-08-17
Add DeleteWikiFinalizeJob

toban:
Adds a job that deletes the remainders of the wiki.

To be run when everthing else has been removed.


app/Jobs/DeleteWikiFinalizeJob.php
```diff
@@ -0,0 +1,82 @@
...
+use App\WikiManager;
+use App\QueryserviceNamespace;
+use App\WikiDb;
+use Traversable;
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Support\Facades\Storage;
+use App\Http\Controllers\WikiLogoController;
+use Illuminate\Contracts\Filesystem\Cloud;
+
+class DeleteWikiFinalizeJob extends Job
```
toban, line 15: implements ShouldBeUnique

toban: done

app/Jobs/DeleteWikiFinalizeJob.php
```diff
@@ -0,0 +1,93 @@
...
+    public function handle()
+    {
+        $wiki = Wiki::withTrashed()->where('id', $this->wikiId )->first();
+
+        if( !$wiki->deleted_at ) {
+            $this->fail(new \RuntimeException("Wiki {$this->wikiId} is not deleted, but job got dispatched."));
+            return;
+        }
+
+        if( !$wiki ) {
```
addshore, line 50: This should go first I think

app/Jobs/DeleteWikiFinalizeJob.php
```diff
@@ -0,0 +1,93 @@
...
+        $wikiDB = $wiki->wikiDb()->first();
+
+
+        if( $wikiDB || $qsNamespace ) {
+            $this->fail(new \RuntimeException("There are still resources allocated."));
+            return;
+        }
+        
+        // if no settings there shouldn't be a need
+        $needToDeleteLogosDir = $wiki->settings()->whereIn('name', array(WikiSetting::wgFavicon, WikiSetting::wgLogo))->count() > 0;
```
addshore, line 65: This should probably change to needsToDeleteSiteDir or something like that.

`sites/'.$siteDir`.

There is probably nota  directory unless people upload a logo right now,
so I imagine the logic would be something like:
 - If site dir exists, delete it, if not don't worry.

tests/Jobs/DeleteWikiFinalizeJobTest.php
```diff
@@ -0,0 +1,107 @@
...
+        $job = new DeleteWikiFinalizeJob( $wiki->id );
+        $job->handle();
+
+        // deletion happened
+        $this->assertNull( Wiki::withTrashed()->where('id', $wiki->id)->first() );
+        $this->assertNull( WikiManager::whereId($manager->id)->first() );
+        $this->assertNull( WikiSetting::whereId($setting->id)->first() );
+
+        // logos dir gone
+        Storage::disk('gcs-public-static')->assertMissing($logoDir);
```
addshore, line 105: Probably want to check for the whole site dir here

addshore (changes requested):

addshore:
This is interesting having throught about elasticsearch a bit more now, as if we delete the wiki_db entry by chance before we delete the ES indexes, then we might never end up cleaning them up D:

toban:
> This is interesting having throught about elasticsearch a bit more now, as if we delete the wiki_db entry by chance before we delete the ES indexes, then we might never end up cleaning them up D:

So, we could do some trickery with events and observers maybe but I'm not so sure you can prevent the actual deletion, and those events don't fire on bulk updates afaik. 

Otherwise maybe adding some model to track the whole ElasticSearch process could work, with that we could also kind of easily solve [the problem with really eager users](https://mattermost.wikimedia.de/swe/pl/xu8zb6dy5bdh9jb3c9yx5mhrbr) by tracking the status of the job a bit better and using `$wgReadOnly` to disable edits before the index is available. Maybe this is what we should have done in the first place. Would require some seeding of existing wikis with the setting enabled once we migrate.

app/Jobs/DeleteWikiFinalizeJob.php
```diff
@@ -0,0 +1,91 @@
...
+        if( $wikiDB || $qsNamespace ) {
+            $this->fail(new \RuntimeException("There are still resources allocated."));
+            return;
+        }
+
+        if ( !$this->deleteSiteDirectory( $wiki->id) ) {
+            $this->fail(new \RuntimeException("Failed deleting logos directory."));
+            return;
+        }
+
```
addshore, line 69: It would be nice to check if elasticsearch index is also deleted here, before we delete the wiki_id which is the reference to it.
I imagine this would look similar to the check in `app/Jobs/ElasticSearchIndexDelete.php` right now to see if indexes exist?

addshore (changes requested):
Looks good other than the 1 final check that it would be nice to have.

app/Jobs/DeleteWikiFinalizeJob.php
toban: added in https://github.com/wbstack/api/pull/158/commits/d8ce639e38f1a1b4fe9eae59ece05efa1fc1b984

PR #161, opened 2021-08-18
[STANDALONE ]Add K8s ingress deletion job

toban:
This turned out to be quite the blackbox, tried setting up minikube for
testing but wouldn't work without somekind of proxy thing to make them
talk between docker <-> minikube which feels to be a step too far. Do we
just trust this api to work?

### DIFF
--- a/tests/Jobs/KubernetesIngressDeleteJobTest.php
+++ b/tests/Jobs/KubernetesIngressDeleteJobTest.php
@@ -1,0 +1,38 @@
+<?php
+
+namespace Tests\Jobs;
+
+use App\User;
+use Illuminate\Foundation\Testing\DatabaseTransactions;
+use Tests\TestCase;
+use App\Jobs\KubernetesIngressDeleteJob;
+use App\Wiki;
+use App\WikiManager;
+use Illuminate\Contracts\Queue\Job;
+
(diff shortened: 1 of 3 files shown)


tests/Jobs/KubernetesIngressDeleteJobTest.php
```diff
@@ -0,0 +1,38 @@
...
+    use DatabaseTransactions;
+
+    public function testDoesNotDeleteNonDeletedWikiws()
+    {
+     
+        $user = User::factory()->create(['verified' => true]);
+        $wiki = Wiki::factory()->create( [ 'deleted_at' => null ] );
+        WikiManager::factory()->create(['wiki_id' => $wiki->id, 'user_id' => $user->id]);
+        
+        $mockJob = $this->createMock(Job::class);
```
toban, line 24: Is this how you are supposed to test these failure functions? It seems to work but this is the first time any of these failure things get tested afaik.

toban: Tried getting the `FailedJobProviderInterface`by using the app->make function which seemed to work, however in the tests nothing seems to get written to the database(also tried adding the Dispatchable interface to the job and use dispatchSync). Does that happen in production though?

addshore (approved):
LGTM

PR #243, opened 2021-11-05
Start using HttpRequest in MediaWikiInit

toban:
- Adds check for response being an array
- Adds test for that case


toban:
tried it out locally and it seems to still work

PR #254, opened 2021-11-22
Add artisan command to soft delete a wiki

jakobw:
Bug: T296180


tarrow (approved):
Code looks awesome; tests are great; I also, after huge quantities of faff, managed to test this thing actually worked locally in my cluster.

PR #266, opened 2021-11-25
Prepare MW 1.37 FP DB upgrade

deer-wmde:
https://phabricator.wikimedia.org/T296178


addshore:
image

I see these insert intos that should not be here.

I think the README step of `Deselect ALL DATA except "content_models", "interwiki", "slot_roles", "updatelog"` might have been missed!

Or rather it looks like this might have been done the wrong way around as I see the inserts for interwiki being removed

image

Generally speaking the rest of it looks great!

deer-wmde:
> Or rather it looks like this might have been done the wrong way around as I see the inserts for interwiki being removed

That could totally be the case, maybe I got it backwards

deer-wmde:
Redid the SQL dump, diff (between the previous 1.36 and the new 1.37 one) can be seen here: https://www.diffchecker.com/J6QTTAI7

tarrow:
I'm not quite sure I understand why this PR adds `database/mw/new/mw1.37-fp-wbs1.sql` but doesn't seem to use it? Maybe wbs2 should become wbs1?

deer-wmde:
> I'm not quite sure I understand why this PR adds `database/mw/new/mw1.37-fp-wbs1.sql` but doesn't seem to use it? Maybe wbs2 should become wbs1?

Indeed!

addshore:
So re reviewing this, it looks like the full list of changes is:

image

image

image

image

image

image

image

Changes to updatelog

addshore (approved):

PR #300, opened 2022-01-07
Add DeleteWikiDispatcherJob.php

toban:
This job schedules multiple separate jobs that reclaim resources for
deleted wikis.

This can be tried out locally by running `skaffold dev` after checking out the dependencies

depends on 

https://github.com/wbstack/charts/pull/43
https://github.com/wmde/wbaas-deploy/pull/137


toban:
Currently this still leaves a row in `wiki_domains` that blocks new wikis to use that domain, we probably want to remove that too?

addshore:
> Currently this still leaves a row in `wiki_domains` that blocks new wikis to use that domain, we probably want to remove that too?

FOr now let's leave it, we would need to ask about the product implications of this.
I could imagine a desired future would be users that delete wikis can reuse those domains, but others can not, or something like that.

app/Jobs/DeleteWikiDispatcherJob.php
```diff
@@ -0,0 +1,83 @@
...
+     */
+    public function handle()
+    {
+        $deleteCutoff = Carbon::now()->subDays(
+            Config::get('wbstack.wiki_hard_delete_threshold')
+        );
+        $wikis = Wiki::withTrashed()->whereDate( 'deleted_at', '<=', $deleteCutoff )->get();
+
+        if( !$wikis->count() ) {
+            Log::info( __METHOD__ . ": Found no soft deleted wikis. exiting.");
```
addshore, line 28:
```suggestion
            Log::info( __METHOD__ . ": Found no soft deleted wikis over threshold. exiting.");
```

app/Jobs/DeleteWikiDispatcherJob.php
```diff
@@ -0,0 +1,83 @@
...
+        foreach($wikis as $wiki) {
+            Log::info( __METHOD__ . ": Starting hard-delete chain for id: {$wiki->id}, domain: {$wiki->domain}.");
+
+            $jobs = [];
+
+            $namespace = QueryserviceNamespace::whereWikiId($wiki->id)->first();
+
+            // if there is a namespace add the job for that
+            if( $namespace ) {
+                $jobs[] = new DeleteQueryserviceNamespaceJob( $namespace->id );
```
addshore, line 41: This is passing the namespace->, but the job actually takes a wiki id here!

addshore, line 41:
```suggestion
                $jobs[] = new DeleteQueryserviceNamespaceJob( $wiki->id );
```

addshore:
I tested the PR with these changes and it seems to work pretty well! :)

app/Jobs/DeleteWikiDispatcherJob.php
toban: geez

toban: good spot, thanks

toban:
Last change might need a bump in some test that checks the logs.

app/Jobs/DeleteWikiDispatcherJob.php
toban: Should probably consider adding some more tests to the QS namespace deletion that it deletes the correct one if there are multiple. or atleast figure out why that slipped.

toban:
CI isn't running :facepalm:

app/Jobs/DeleteWikiDispatcherJob.php
addshore: This would be one reason to argue FOR typing / passing an object here, and against just ints

addshore:
Rebased to get rid of composer.lock conflict

addshore (approved):
Looks good, and works well in conjunction with the other 2 PRs

toban:
squash and merge me

tests/Jobs/DeleteWikiDispatcherJobTest.php
```diff
@@ -86,7 +94,7 @@ public function testNothingDispatchesUntilItsTime() {
         $job->handle();
 
         Log::assertLogged('info', function ($message, $context) {
-            return Str::contains($message, 'Found no soft deleted wikis. exiting.');
+            return Str::contains($message, 'Found no soft deleted wikis over threshold. exiting.');
```
addshore, line 97: Aaah sorry, I totally broke this!

PR #301, opened 2022-01-07
Enable setting a read only setting for a wiki

addshore:
This also allows the set setting command for a wiki
to be able to set it to blank / delete the entry.
This is ONLY setable internally by the platform, and not via the API.

The command would be:
php artisan wbs-wiki:setSetting

And you can use something like the following to set readonly:
domain wiki1.wbaas.localhost wgReadOnly "Message here"

Or this to unset:
domain wiki1.wbaas.localhost wgReadOnly

This is for https://phabricator.wikimedia.org/T298759
and also has a MediaWiki PR for the config
https://github.com/wbstack/mediawiki/pull/197

### DIFF
--- a/app/Console/Commands/Wiki/SetSetting.php
+++ b/app/Console/Commands/Wiki/SetSetting.php
@@ -8,7 +8,7 @@ use Illuminate\Console\Command;
 
 class SetSetting extends Command
 {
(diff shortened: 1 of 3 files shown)


jakobw (approved):
Works beautifully :tada: 

Screenshot from 2022-01-10 11-43-19

PR #309, opened 2022-01-11
Update README.md

outdooracorn:


jakobw (commented):
I think the idea is for local development to not happen on metal, but through docker-compose. These dependencies including the composer ones are already built into the image.

toban:
> I think the idea is for local development to not happen on metal, but through docker-compose. These dependencies including the composer ones are already built into the image.

I agree, lets maybe write that in the header or something that this would only be required if we are running outside of the docker dev environment not to lure anyone in thinking this is an actual dependency to be installed to run the api.

toban (changes requested):
Lets add a note that this is only required if running outside of docker.

outdooracorn:
Thanks for the reviews. I think you are right that the preferred way to run `composer` is in a container. At the time of writing this PR the docs simply said to run `composer install` (i.e. run it locally) which does require the php extensions to be installed first.

The docs have since been updated to include a way to do this with docker.

I have updated the docs in this PR to clarify that docker is the preferred way of doing this but if you want to run `composer` locally you will need the `xml` and `curl` php extensions.

README.md
```diff
@@ -14,10 +14,17 @@ and query service management code.
...
-`composer install`
+It is recommened to install the php dependencies via a composer docker container as that contains the required php extensions: 
+```
+docker run --rm -it -v $PWD:/app -u $(id -u):$(id -g) composer install
+```
+
 
-Alternatively you can use docker to run composer
-`docker run --rm --interactive --tty --volume $PWD:/app --user $(id -u):$(id -g) composer install`
+Alternatively, you can run `composer` on your local machine but make sure the `xml` and `curl` php extensions are installed. For example, on Ubuntu:
```
jakobw, line 23: If the intention here is to run composer on metal and build the dependencies to be used within a container, the best way is probably to just tell it to ignore any extension or language version requirements.

```suggestion
Alternatively, you can run `composer` on your local machine (ignoring platform requirements):
```

I think ideally we would want a version of the API image that has composer built into it, so that composer uses the same PHP version and environment as the application code. Then we could run it with something like `docker-compose run --rm api composer install`, but that doesn't have to happen now.

README.md
```diff
@@ -14,10 +14,17 @@ and query service management code.
...
+```
+sudo apt install php-xml php-curl
+composer install
```
jakobw, line 26:
```suggestion
composer install --ignore-platform-reqs
```

(Funny, I had to split this up into two comments, because apparently GH can't handle suggestions that include triple backticks <code>```</code>)

jakobw (changes requested):

README.md
outdooracorn: > I think ideally we would want a version of the API image that has composer built into it, so that composer uses the same PHP version and environment as the application code. Then we could run it with something like `docker-compose run --rm api composer install`, but that doesn't have to happen now.

Cool, that makes sense. In the interim, I noticed that the `composer.json` specifies the PHP version (7.4) which I think helps to ensure the same PHP version is used (environment not so much).
https://github.com/wbstack/api/blob/f29860703fddc3529bf38bc300f7068c3d72efb1/composer.json#L75-L78

README.md
outdooracorn: Cool, thanks for pointing out that `composer` flag

README.md
jakobw: Yes, but one of the features of `--ignore-platform-reqs` is ignoring this line :|

PR #318, opened 2022-01-12
Add Makefile

toban:
I've had bash-script todo this for a while but never comitted it
anywhere. Lets add a Makefile where we can add common things we want to
run for this repo.


Makefile
```diff
@@ -0,0 +1,11 @@
+#shellcheck disable=code
+.PHONY: test-and-migrate migrate test
+
+# Example: make test FILTER=tests/Jobs/ElasticSearchIndexInitTest.php
+test:
+	docker-compose exec api vendor/bin/phpunit ${FILTER}
+
+migrate:
```
jakobw, line 8: Hmm, `migrate` isn't an ideal name here IMO. Running laravel migrations is just `php artisan migrate`, and this long command here is more of an initial db setup kind of thing?

jakobw (changes requested):
Good idea in general, but not super happy with the "migrate" naming.

Makefile
addshore: +1

toban: `init` instead?

Makefile
```diff
@@ -0,0 +1,11 @@
...
+	docker-compose exec api bash -c 'php artisan migrate:fresh && php artisan passport:install && php artisan db:seed && php artisan key:generate'
+
+test-and-migrate: migrate test
```
toban, line 11:
```suggestion
test-fresh: init test
```

Makefile
```diff
@@ -0,0 +1,11 @@
+#shellcheck disable=code
+.PHONY: test-and-migrate migrate test
```
toban, line 2:
```suggestion
.PHONY: test-fresh init test
```

jakobw (approved):
:rocket:

Issue #317, opened 2022-01-12
wiki domain seed not random enough


toban:
From time to time CI fails because of domains seems to be getting re-used

```
1) Tests\Jobs\DeleteWikiDispatcherJobTest::testNothingDispatchesUntilItsTime
Illuminate\Database\QueryException: SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry 'harvey.com' for key 'wikis_domain_unique' (SQL: insert into `wikis` (`sitename`, `domain`, `deleted_at`, `updated_at`, `created_at`) values (Non., harvey.com, ?, 2022-01-12 08:44:04, 2022-01-12 08:44:04))

/var/www/html/vendor/laravel/framework/src/Illuminate/Database/Connection.php:705
/var/www/html/vendor/laravel/framework/src/Illuminate/Database/Connection.php:665
/var/www/html/vendor/laravel/framework/src/Illuminate/Database/Connection.php:495
/var/www/html/vendor/laravel/framework/src/Illuminate/Database/Connection.php:447
/var/www/html/vendor/laravel/framework/src/Illuminate/Database/Query/Processors/Processor.php:32
/var/www/html/vendor/laravel/framework/src/Illuminate/Database/Query/Builder.php:2966
/var/www/html/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Builder.php:1655
/var/www/html/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php:1188
/var/www/html/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php:1153
/var/www/html/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php:994
/var/www/html/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Factories/Factory.php:314
/var/www/html/vendor/laravel/framework/src/Illuminate/Collections/Traits/EnumeratesValues.php:242
/var/www/html/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Factories/Factory.php:317
/var/www/html/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Factories/Factory.php:261
/var/www/html/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Factories/Factory.php:255
/var/www/html/tests/Jobs/DeleteWikiDispatcherJobTest.php:43
/var/www/html/tests/Jobs/DeleteWikiDispatcherJobTest.php:80
[...]

PR #322, opened 2022-01-13
Fix WikiFactory

jakobw:
The faker in `ModelFactory` was not used.

There was a suspicious failure in #309 (https://github.com/wbstack/api/runs/4805489685?check_suite_focus=true) and it turns out that the `WikiFactory` wasn't using the definition as expected. I haven't checked the other factories but looking at the docs this seems to have changed with the latest major laravel version:
https://laravel.com/docs/8.x/database-testing#defining-model-factories
https://laravel.com/docs/7.x/database-testing#writing-factories

### DIFF
--- a/database/factories/ModelFactory.php
+++ b/database/factories/ModelFactory.php
@@ -25,14 +25,6 @@ $factory->define(App\Invitation::class, function (Faker\Generator $faker) {
     ];
 });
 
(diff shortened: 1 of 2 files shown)


toban:
Nice looks like this could be the fix for https://github.com/wbstack/api/issues/317

toban:
I remember this being something being worked on when doing the laravel update, looks like it wasn't done right.

PR #331, opened 2022-01-19
Add jobs to run CirrusSearch ForceSearchIndex

toban:
Adds jobs for queuing batched runs of ForceSearchIndex.php

Could probably be refactored to use https://laravel.com/docs/8.x/queues#job-batching at some point but this felt like overkill now and would make this pull-request too big.

Timeouts was chosen a bit at random, I think we need to try running this on the cloud provider first to find out what is reasonable.

depends on https://github.com/wbstack/mediawiki/pull/209


app/Http/Controllers/WikiController.php
```diff
@@ -6,7 +6,7 @@
 use App\Jobs\MediawikiInit;
 use App\Jobs\ProvisionQueryserviceNamespaceJob;
 use App\Jobs\ProvisionWikiDbJob;
-use App\Jobs\ElasticSearchIndexInit;
+use App\Jobs\ElasticSearch\ElasticSearchIndexInit;
```
toban, line 9: This should probably say CirrusSearch instead of ElasticSearch as this is what it's interacting with.

tests/Jobs/ElasticSearch/ElasticSearchIndexDeleteTest.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-namespace Tests\Jobs;
+namespace Tests\Jobs\ElasticSearch;
```
toban, line 3: This might be incorrect if the namespace changes to CirrusSearch, this job actually interacts directly with ElasticSearch.

app/Http/Controllers/WikiController.php
addshore: Sounds good to me!

app/Jobs/ElasticSearch/ForceSearchIndex.php
```diff
@@ -0,0 +1,66 @@
...
+ * Example:
+ * 
+ * php artisan job:dispatchNow ElasticSearch\\ForceSearchIndex 1 0 1000 ,
+ */
+class ForceSearchIndex extends CirrusSearchJob
+{
+    private $fromId;
+    private $toId;
+
+    public function __construct( int $wikiId, int $fromId, int $toId  ) {
```
addshore, line 20: It would be great to make this consistent with some of the other jobs that allow selecting by domain etc.
This will especially be handy for the migration

app/Jobs/ElasticSearch/ForceSearchIndex.php
```diff
@@ -0,0 +1,66 @@
...
+
+use Illuminate\Support\Facades\Log;
+
+/**
+ * 
+ * Job for running CirrusSearch/ForceSearchIndex.php on a wiki
+ * 
+ * Example:
+ * 
+ * php artisan job:dispatchNow ElasticSearch\\ForceSearchIndex 1 0 1000 ,
```
addshore, line 13:
```suggestion
 * php artisan job:dispatchNow ElasticSearch\\ForceSearchIndex 1 0 1000
```

app/Jobs/ElasticSearch/QueueSearchIndexBatches.php
```diff
@@ -0,0 +1,86 @@
...
+use Illuminate\Support\Facades\Log;
+use Illuminate\Foundation\Bus\DispatchesJobs;
+
+/**
+ * 
+ * Job for queuing batched runs of CirrusSearch ForceSearchIndex.php by using the --buildChunks parameter 
+ * 
+ * Example:
+ * 
+ * php artisan job:dispatchNow ElasticSearch\\QueueSearchIndexBatches 1 ,
```
addshore, line 15:
```suggestion
 * php artisan job:dispatchNow ElasticSearch\\QueueSearchIndexBatches 1
```

tests/Jobs/ElasticSearch/ElasticSearchIndexDeleteTest.php
addshore: Perhaps this one should remain in ElasticSearch then?

toban: yup

app/Jobs/ElasticSearch/ForceSearchIndex.php
toban: done

PR #334, opened 2022-01-20
Update to dev `fakerphp/faker` & user composer 2 in Dockerfile

outdooracorn:


addshore:
Using composer 2 in the dockerfile will also unblock https://github.com/wbstack/api/pull/328

addshore:
I failed to spot that this change not only switches to composer 2 and `fakerphp/faker` but also does a `composer update`
This will lead to a bunch of dependencies updating in a way where we are not really looking, so I'm going to rollback the composer.lock change, and only add the new faker package.

addshore:
https://github.com/wbstack/api/pull/337

PR #335, opened 2022-01-20
Do some database cleaning before tests

rosalieper:
- Delete database entries created during test
- Remove trialling spaces


addshore (changes requested):
SO I don't think we want to merge this sort of cleanup.
These tests should create things that they will use for testing, and `DatabaseTransactions` should ensure that anything written by this test is removed again once it is completed.

PR #339, opened 2022-01-21
Add seperate Xdebug configuration

deer-wmde:
This PR adds some separate configuration to use Xdebug when developing with docker-compose.

There's a separate Dockerfile, which installs and configures Xdebug, and a docker-compose file which uses that Dockerfile, plus adding the docker host for the container.

To run the container in 'xdebug mode' I added the `xdebug` command to our Makefile, which builds and runs the container with the configuration above. Once that is running you can use `make test` and you should be able to use Xdebug with your IDE :tada: (note: I only tested this with PhpStorm)

If we agree that we want this in the repo I'll add some docs before merging, because the usage isnt quite obvious I think.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,3 +12,10 @@ test-fresh: init test
 
 lint:
 	docker-compose exec -T api vendor/bin/psalm
+
+# Example: make test-xdebug FILTER=tests/Jobs/ElasticSearchIndexInitTest.php
+test-xdebug:
+	docker-compose exec api bash -c 'PHP_IDE_CONFIG="serverName=wbaas-local-api" XDEBUG_SESSION=1 php vendor/bin/phpunit ${FILTER}'
+
+xdebug:
+	docker-compose -f docker-compose.debug.yml up -d
--- a/Makefile
+++ b/Makefile
@@ -13,9 +13,5 @@ test-fresh: init test
 lint:
 	docker-compose exec -T api vendor/bin/psalm
 
-# Example: make test-xdebug FILTER=tests/Jobs/ElasticSearchIndexInitTest.php
-test-xdebug:
-	docker-compose exec api bash -c 'PHP_IDE_CONFIG="serverName=wbaas-local-api" XDEBUG_SESSION=1 php vendor/bin/phpunit ${FILTER}'
-
 xdebug:
 	docker-compose -f docker-compose.debug.yml up -d
--- a/debug.Dockerfile
+++ b/debug.Dockerfile
@@ -1,0 +1,56 @@
+FROM composer@sha256:d374b2e1f715621e9d9929575d6b35b11cf4a6dc237d4a08f2e6d1611f534675 as composer
+# composer is pinned at a PHP 7 version
+
+COPY ./composer.json /tmp/src1/composer.json
+COPY ./composer.lock /tmp/src1/composer.lock
+
+WORKDIR /tmp/src1
+RUN composer install --no-dev --no-progress --no-autoloader
+
+COPY ./ /tmp/src2
+RUN cp -r /tmp/src1/vendor /tmp/src2/vendor
+WORKDIR /tmp/src2
+RUN composer install --no-dev --no-progress --optimize-autoloader
+
+
+FROM php:7.4-apache
+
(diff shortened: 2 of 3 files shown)


deer-wmde:
also I'd be super happy if someone knows a way to simplify the Dockerfile and docker-compose file I copied. Not sure if its possible to reuse the existing ones somehow (without copying them)

deer-wmde:
Ideas for another/slimmer approach:

Are we using the docker-compose file only for local development? If that is the case, we could add the important line for the xdebug connection to the default one and get rid of `docker-compose.debug.yml`. Maybe the `xdebug` Makefile command then could install and configure xdebug :thinking:

toban:
> Ideas for another/slimmer approach:
> 
> Are we using the docker-compose file only for local development? If that is the case, we could add the important line for the xdebug connection to the default one and get rid of `docker-compose.debug.yml` and `debug.Dockerfile`, if maybe the `xdebug` Makefile command then could install and configure xdebug thinking

Yes only for local development and testing in Github Actions. I think that sounds like a good idea, possibly even mount that file into the container?

debug.Dockerfile
```diff
@@ -0,0 +1,59 @@
+FROM composer@sha256:d374b2e1f715621e9d9929575d6b35b11cf4a6dc237d4a08f2e6d1611f534675 as composer
```
toban, line 1: In general not a super big fan of having two docker images we need to keep in sync, I think this could be solved by reusing the existing docker file but adding the required dependencies and then enabling them by some volume / --build-argument thing possibly?

debug.Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+
+RUN pecl install xdebug \
+    && docker-php-ext-enable xdebug
+
+RUN echo "zend_extension=xdebug\n\n\
+[xdebug]\n\
+xdebug.mode=develop,debug\n\
+xdebug.client_host=host.docker.internal\n\
+xdebug.client_port=9003\n\
+" > /usr/local/etc/php/conf.d/docker-php-ext-xdebug.ini
```
toban, line 40: This file could be mounted into the docker-compose.yml file used for development

docker-compose.debug.yml
```diff
@@ -0,0 +1,50 @@
+version: '2'
```
toban, line 1: Same here, not a big fan of copy and pasting this, we should probably re-use the existing one as it is only used for development / testing. Worst-case we could add an [docker-compose.override.yml](https://docs.docker.com/compose/extends/) that just adds the modifications we want, but I don't see any reason not to try to get xdebug working for the existing local development configuration.

Makefile
```diff
@@ -12,3 +12,6 @@ test-fresh: init test
 
 lint:
 	docker-compose exec -T api vendor/bin/psalm
+
+xdebug:
```
toban, line 16: I think we should have this named `start` or something like that rather than xdebug, we want xdebug to work for the local development environment.

toban (commented):
Some thoughts that came up looking through this PR. 

Thanks for working on this it will help us a lot going forward.

Would possibly also be worth adding some notes / links in a readme somewhere on how you got it working in the IDE too. Maybe a link / copy of an updated version of that gist you had earlier?

debug.Dockerfile
toban: I've done something similar to what you want to do (i think) in the release pipeline where a --build-argument is used as a conditional to either install dependencies or not.

https://github.com/wmde/wikibase-release-pipeline/blob/main/Docker/test/selenium/Dockerfile#L13

toban: This was then triggered on build by adding it here https://github.com/wmde/wikibase-release-pipeline/blob/main/test/run_selenium.sh#L40 Maybe something like that could work?

outdooracorn: The alternative is to build the debug image from the production image (e.g. `FROM ghcr.io/wbstack/api`) then you don't need to copy pasta everything over and keep in sync. Not sure if which method I prefer though.

toban:
added a phab task for this work here https://phabricator.wikimedia.org/T300117

PR #346, opened 2022-01-25
Specifically require http/socket-client:2.0.2

tarrow:
http/socket-client:2.0.1 has a bug becuase it claims to provide
psr/http-client yet it doesn't

We struggled to update this with `composer update` so for now
we specifically require this version


addshore:
*rage*, but I think this is what we want to do for now

PR #348, opened 2022-01-25
Fix logo path for resizing

addshore:


toban (approved):
Poked around a bit with this locally and this still seems to work fine with the test and is correctly using the output of the job when testing. Seems Image::make can take a whole range of things to produce the image.

removed the unused rawPath

outdooracorn (approved):
I also, tested locally and works fine! :shipit:

PR #349, opened 2022-01-25
Add `auth.json` to `.gitignore` file

outdooracorn:
`auth.json` stores repository (e.g. github) tokens for use with `composer` to avoid being rate limited


toban (approved):
works like a charm

PR #355, opened 2022-01-26
Add checks for MW api errors in CirrusSearchJob

toban:
This add checks for errors that come from the Mediawiki API in CirrusSearchJobs, for example if a certain module isn't
 available or some database error this would get caught.


rosalieper (approved):
This look good to me. A nice to have.

tarrow (approved):

PR #364, opened 2022-02-03
Handle custom domains in MigrateWikiCreate Job

outdooracorn:


app/Jobs/MigrationWikiCreate.php
```diff
@@ -105,6 +106,12 @@ private function createWiki($user, $wikiDetails, $wikiDb) {
             $wikiDomain->wiki_id = $wiki->id;
             $wikiDomain->save();
 
+            // Create k8s ingress if the wiki uses a custom domain
+            if ( ! WikiController::isSubDomain($wikiDetails->domain)) {
+                // TODO: should this be done synchronously with dispatch_sync(), or asynchronously with dispatch()?
```
toban, line 111: Couldn't we do this outside of transaction to only happen if the whole thing is successful? I don't think in we'd need to do it in sync with this job then either? looks like you could use `$wiki->domain` if set maybe?

toban (commented):
Some thoughts that came up when i looked through the patch.

app/Jobs/MigrationWikiCreate.php
```diff
@@ -111,6 +112,11 @@ private function createWiki($user, $wikiDetails, $wikiDb) {
             ]);
         });
 
+        // Create k8s ingress if the wiki uses a custom domain
+        if ( isset($wiki->domain) && ! WikiController::isSubDomain($wiki->domain)) {
```
toban, line 116: Cool, this first isset to me looks like a more serious failure that might warrant marking the job as a failure if that happens. If we get down here and the $wiki object is incomplete that probably means the transaction failed for some reason but the migration job would still be considered valid? I guess ultimately this depends on how we run these things. If we don't see it happening in some controller at some point but rather from the command line every time, maybe this doesn't matter?

toban (commented):
LGTM, one comment/question that I could live without also depending on how we see this running, what do you think?

PR #404, opened 2022-04-01
Update email copy

toban:
This can be tried out by running `skaffold -m api` locally and triggering these three emails.


resources/views/vendor/notifications/email.blade.php
```diff
@@ -0,0 +1,62 @@
+@component('mail::message')
+{{-- Greeting --}}
+@if (! empty($greeting))
+# {{ $greeting }}
+@else
+@if ($level === 'error')
+# @lang('Whoops!')
+@else
+# @lang('Hello!')
```
toban, line 9: Added a question on this being left in or not in the google doc describing the new copy.

deer-wmde (approved):
I tested this locally and I approve. I appreciate moving the mail templates into the repo - not sure if its needed here but it probably will in the future! :tada:

PR #429, opened 2022-05-04
Remove unused data from /mine

toban:
This just makes the initial load slow and doesn't seem to be used anywhere.


deer-wmde:
tested it locally and to me it looks like it works as it is :ok_hand:

PR #460, opened 2022-06-16
Do not depend on previous main builds

rosalieper:


toban:
LGTM seems to work well when developing locally and updating the code to running tests again too.

PR #497, opened 2022-08-03
Refactor stats summary

toban:
output to terminal instead of email
add version
remove notification class
only output outside of tests


app/Jobs/PlatformStatsSummaryJob.php
```diff
@@ -29,6 +30,8 @@
 class PlatformStatsSummaryJob extends Job
 {
     private $inactiveThreshold;
+
+    private $platformSummaryStatsVersion = "v1";
     protected $email;
```
tarrow, line 35: I guess this field should now be removed?

tarrow (approved):
Looks good to me, would be nice to remove the unused email field though

app/Jobs/PlatformStatsSummaryJob.php
toban: yar!

toban: done

deer-wmde (approved):
Looks great!

PR #506, opened 2022-08-15
Add active users to PlatformStatsSummaryJob

deer-wmde:
This PR adds active users count to PlatformStatsSummaryJob https://phabricator.wikimedia.org/T314128


tarrow (commented):
Looks good, also seems to correct type for the test case wikis

PR #507, opened 2022-08-16
Disable elasticsearch for new wikis by default

toban:
This adds the config parameter `elasticsearch_enabled_by_default`
to control if the feature is enabled for newly created wikis or not.

This also refactors one test to make this support both behaviors.

Bug: T314937


rosalieper:
this looks good

rosalieper (approved):

PR #509, opened 2022-08-18
Refactor /mine to include wikis, the count and the limit

toban:
Bug: T315441


app/Http/Controllers/WikisController.php
```diff
@@ -9,10 +9,14 @@ class WikisController extends Controller
...
-          $request->user()
-          ->managesWikis()
-          ->get()
-        );
+      $wikis = $request->user()->managesWikis()->get();
+
+      return response(
+        [
+          "wikis" => $wikis,
+          "count" => count($wikis),
```
toban, line 17: One could argue that this is useless.

rosalieper (approved):

PR #586, opened 2023-03-10
Remove MediawikiManualDbUpdate and references

tarrow:
This legacy update method had been replace by a job that goes on to trigger update.php.

This commit removes references to the old methods to reduce confusion


deer-wmde (approved):
I approve. This looks like legacy wbstack things that probably weren't used in the last 2 years and I don't find any usages in our repos for the removed Jobs.

Thanks for cleaning up 🧹

PR #589, opened 2023-03-16
Update database scripts to support MediaWiki 1.38

AndrewKostka:


database/mw/docker-compose-clean.yml
```diff
@@ -3,9 +3,10 @@ version: '3.1'
 services:
 
   mediawiki:
-    image: ghcr.io/wbstack/mediawiki:main
+    image: ghcr.io/wbstack/mediawiki:1.38-7.4-20230316-0
     volumes:
       - ./maintWikWiki.json:/var/www/html/w/maintWikWiki.json
+      - ./:/usr/src/
```
m90, line 9: I wonder if this mount accidentally overwrites anything the image expects to be there?

Also, `man hier` advises against putting your own stuff in `/usr/src`:

```
       /usr/src
              Source files for different parts of the system, included with some packages for reference purposes.  Don't work here with your own projects, as files below /usr  should  be  read-only
              except when installing software (optional).
```

Would it be a thing to just mount the php script in `maintenance` like this:

```
- ./cleanSql.php:/var/www/html/w/maintenance/cleanSql.php
```

m90 (approved):
I left one very very bikesheddy comment which you can disregard if you feel like it. Everything else looks good to me.

deer-wmde:
The DB dump seems good to me! https://www.diffchecker.com/mPYZnS3j/

database/mw/docker-compose-clean.yml
AndrewKostka: Good idea, thanks!

deer-wmde (approved):

PR #590, opened 2023-03-16
Remove unused MediaWiki DB migration scripts

AndrewKostka:


addshore:
I support this if update.php will continue to be used moving forward

deer-wmde:
I think in the meantime these changes made it into main via #586

PR #595, opened 2023-04-03
Do not disable elastic search on wikis after a failure

rosalieper:
Also remove some trailing spaces
Bug: T333559

### DIFF
--- a/app/Jobs/CirrusSearch/CirrusSearchJob.php
+++ b/app/Jobs/CirrusSearch/CirrusSearchJob.php
@@ -59,10 +59,9 @@ abstract class CirrusSearchJob extends Job implements ShouldBeUnique
         // no wikiDB around
         if ( !$this->wikiDB ) {
             $this->fail( new \RuntimeException($this->apiModule() . ' call for '.$this->wikiId.' was triggered but not WikiDb available') );
-            $this->setting->update( [  'value' => false  ] );
             return;
         }
-        
+
         $request->setOptions(
             [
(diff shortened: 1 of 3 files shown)


tarrow:
I *think* these failures in the tests may be all totally expected and simply require updating the tests.

It looks to me like the tests are checking each time at elasticsearch is disabled (see: https://github.com/wbstack/api/blob/main/tests/Jobs/CirrusSearch/ElasticSearchIndexInitTest.php#L161) given that we've just stopped this from happening we probably also need to update the tests.

I guess we might want to assert something else (for example that the logger is correctly called) instead.

deer-wmde:
The code looks pretty good to me but I still want to test it in my local cluster, so far I've only read the code and ran the unit tests.

And I saw there was a lot of work put into making the phpunit test work, but I think there wasn't much to change in the first place to make it work :D I'm not sure if all the changes were necessary as the test passes if we just change the expected setting count value from 0 to 1 (which means: instead of expecting none (0), we now expect there to be one (1) WikiSetting existing, because we create it anyway, even if theres some error)

What do you think @rosalieper ? Happy to chat about it in a call if its not clear what I mean (or maybe I even misunderstood the test code :sweat_smile: )

rosalieper:
yeah. I agree. rather than removing the assert we change the expected value. This will also make the test reflect the current logic.

deer-wmde (approved):
Awesome, I could test this locally and I think it works! great work

tarrow (approved):

PR #600, opened 2023-04-26
Add documentation for ES re-indexing for wikis

deer-wmde:
I thought it might be useful to have some documentation on our preferred way to re-index wikis.

Feel free to discuss if you think that this is not the way we should do it or if you think this documentation should live elsewhere (in the deploy-repo for example).


m90 (approved):
Brilliant, thanks a lot :1st_place_medal:

PR #601, opened 2023-05-03
When wikis have pending jobs, create a Kubernetes job to work through the items

m90:
Ticket https://phabricator.wikimedia.org/T330389

Requires https://github.com/wbstack/charts/pull/114, https://github.com/wmde/wbaas-deploy/pull/885

### DIFF
--- a/app/Console/Kernel.php
+++ b/app/Console/Kernel.php
@@ -43,6 +43,8 @@ class Kernel extends ConsoleKernel
 
         // Schedule site stat updates for each wiki and platform-summary
         $schedule->command('schedule:stats')->daily();
+
+        $schedule->job(new PollForMediaWikiJobsJob)->everyMinute();
     }
 
     /**
--- a/app/Console/Kernel.php
+++ b/app/Console/Kernel.php
@@ -8,6 +8,7 @@ use App\Jobs\ProvisionWikiDbJob;
 use App\Jobs\PruneEventPageUpdatesTable;
 use App\Jobs\PruneQueryserviceBatchesTable;
 use App\Jobs\SandboxCleanupJob;
+use App\Jobs\PollForMediaWikiJobsJob;
 use Illuminate\Console\Scheduling\Schedule;
 use Illuminate\Foundation\Console\Kernel as ConsoleKernel;
 use App\Jobs\PlatformStatsSummaryJob;
--- a/app/Jobs/PollForMediaWikiJobsJob.php
+++ b/app/Jobs/PollForMediaWikiJobsJob.php
@@ -3,6 +3,7 @@
 namespace App\Jobs;
 
 use App\Wiki;
+use Illuminate\Support\Facades\Http;
 
 class PollForMediaWikiJobsJob extends Job
 {
@@ -18,7 +19,23 @@ class PollForMediaWikiJobsJob extends Job
 
     private function hasPendingJobs( string $wikiDomain ): bool
     {
-        return true;
+        $response = Http::withHeaders([
+            'host' => $wikiDomain
+        ])->get(
+            getenv('PLATFORM_MW_BACKEND_HOST').'/w/api.php?action=query&meta=siteinfo&siprop=statistics&format=json'
+        );
+
+        if ($response->failed()) {
+            $this->fail(
+                new \RuntimeException(
+                    'Failure polling wiki '.$wikiDomain.' for pending MediaWiki jobs: '.$response->clientError()
+                )
+            );
+
+            return false;
+        }
+
+        return $response->json()['query']['statistics']['jobs'] > 0;
     }
 
     private function enqueueWiki ( string $wikiDomain ): void
--- a/app/Jobs/PollForMediaWikiJobsJob.php
+++ b/app/Jobs/PollForMediaWikiJobsJob.php
@@ -35,7 +35,8 @@ class PollForMediaWikiJobsJob extends Job
             return false;
         }
 
-        return $response->json()['query']['statistics']['jobs'] > 0;
+        $pendingJobs = data_get($response->json(), 'query.statistics.jobs', 0);
+        return $pendingJobs > 0;
     }
 
     private function enqueueWiki ( string $wikiDomain ): void
--- a/app/Jobs/PollForMediaWikiJobsJob.php
+++ b/app/Jobs/PollForMediaWikiJobsJob.php
@@ -18,7 +18,7 @@ class PollForMediaWikiJobsJob extends Job
 
     private function hasPendingJobs( string $wikiDomain ): bool
     {
-        return false;
+        return true;
     }
 
     private function enqueueWiki ( string $wikiDomain ): void
--- a/app/Jobs/PollForMediaWikiJobsJob.php
+++ b/app/Jobs/PollForMediaWikiJobsJob.php
@@ -8,7 +8,7 @@ class PollForMediaWikiJobsJob extends Job
 {
     public function handle(): void
     {
-        $wikis = Wiki::all()->pluck( 'domain' )->get();
+        $wikis = Wiki::all()->pluck('domain');
         foreach ( $wikis as $wikiDomain ) {
             if ($this->hasPendingJobs( $wikiDomain )) {
                 $this->enqueueWiki( $wikiDomain );
@@ -18,7 +18,7 @@ class PollForMediaWikiJobsJob extends Job
 
     private function hasPendingJobs( string $wikiDomain ): bool
     {
-        return false
+        return false;
     }
 
     private function enqueueWiki ( string $wikiDomain ): void
--- a/app/Jobs/PollForMediaWikiJobsJob.php
+++ b/app/Jobs/PollForMediaWikiJobsJob.php
@@ -1,0 +1,28 @@
+<?php
+
+namespace App\Jobs;
+
+use App\Wiki;
+
+class PollForMediaWikiJobsJob extends Job
+{
+    public function handle(): void
+    {
+        $wikis = Wiki::all()->pluck( 'domain' )->get();
+        foreach ( $wikis as $wikiDomain ) {
+            if ($this->hasPendingJobs( $wikiDomain )) {
+                $this->enqueueWiki( $wikiDomain );
+            }
+        }
+    }
+
+    private function hasPendingJobs( string $wikiDomain ): bool
+    {
+        return false
+    }
+
+    private function enqueueWiki ( string $wikiDomain ): void
+    {
+        dispatch(new ProcessMediaWikiJobsJob( $wikiDomain ));
+    }
+}
--- a/app/Jobs/PollForMediaWikiJobsJob.php
+++ b/app/Jobs/PollForMediaWikiJobsJob.php
@@ -10,14 +10,14 @@ class PollForMediaWikiJobsJob extends Job
     public function handle(): void
     {
         $wikis = Wiki::all()->pluck('domain');
-        foreach ( $wikis as $wikiDomain ) {
-            if ($this->hasPendingJobs( $wikiDomain )) {
-                $this->enqueueWiki( $wikiDomain );
+        foreach ($wikis as $wikiDomain) {
+            if ($this->hasPendingJobs($wikiDomain)) {
+                $this->enqueueWiki($wikiDomain);
             }
         }
     }
 
-    private function hasPendingJobs( string $wikiDomain ): bool
+    private function hasPendingJobs(string $wikiDomain): bool
     {
         $response = Http::withHeaders([
             'host' => $wikiDomain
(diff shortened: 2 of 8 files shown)


app/Console/Kernel.php
```diff
@@ -43,6 +44,8 @@ protected function schedule(Schedule $schedule)
 
         // Schedule site stat updates for each wiki and platform-summary
         $schedule->command('schedule:stats')->daily();
+
+        $schedule->job(new PollForMediaWikiJobsJob)->everyMinute();
```
m90, line 48: As Laravel is using cron schedules for defining job rates, there is no way to have this run more often. Unless we roll our own ticker or similar. Not sure if this frequency is ok for us?

app/Jobs/PollForMediaWikiJobsJob.php
```diff
@@ -0,0 +1,45 @@
...
+    {
+        $wikis = Wiki::all()->pluck('domain');
+        foreach ( $wikis as $wikiDomain ) {
+            if ($this->hasPendingJobs( $wikiDomain )) {
+                $this->enqueueWiki( $wikiDomain );
+            }
+        }
+    }
+
+    private function hasPendingJobs( string $wikiDomain ): bool
```
m90, line 20: I wonder if there is a way to do this with a little less HTTP involved? Or is this the go-to solution for us?

app/Jobs/PollForMediaWikiJobsJob.php
```diff
@@ -0,0 +1,45 @@
...
+            $this->fail(
+                new \RuntimeException(
+                    'Failure polling wiki '.$wikiDomain.' for pending MediaWiki jobs: '.$response->clientError()
+                )
+            );
+
+            return false;
+        }
+
+        return $response->json()['query']['statistics']['jobs'] > 0;
```
m90, line 38: In PHP, is there an idiomatic way of safely accessing a nested array without the possibility of a Runtime Exception thrown on malformed data? I.e. how could I make sure this gracefully handles a json response of `{'error': 'Bad Gateway'}`

app/Jobs/ProcessMediaWikiJobsJob.php
```diff
@@ -0,0 +1,114 @@
...
+    private string $wikiDomain;
+
+    public function __construct ( string $wikiDomain )
+    {
+        $this->wikiDomain = $wikiDomain;
+    }
+
+    public function uniqueId(): string
+    {
+        return $this->wikiDomain;
```
m90, line 26: This makes sure only a single job per wiki domain is ever enqueued. If the scheduler pushes another job for the same wiki onto the queue, it will be dropped instantly.

app/Jobs/ProcessMediaWikiJobsJob.php
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        $mwPod = $mwPod->toArray();
+
+        $jobSpec = new KubernetesJob([
+            'metadata' => [
+                'generateName' => 'run-all-mw-jobs-'
+            ],
+            'spec' => [
+                'ttlSecondsAfterFinished' => 60 * 60,
```
m90, line 55: Is it a sane idea to keep job objects around for an hour only?

app/Jobs/ProcessMediaWikiJobsJob.php
```diff
@@ -0,0 +1,114 @@
...
+                'template' => [
```
m90, line 56: I was wondering whether we should put these jobs into a dedicated namespace. In case yes, who should be responsible for creating the (empty) namespace? This job, or `wbaas-deploy`?

app/Jobs/ProcessMediaWikiJobsJob.php
```diff
@@ -0,0 +1,114 @@
...
+                    'spec' => [
+                        'containers' => [
+                            0 => [
+                                'name' => 'run-all-mw-jobs',
+                                'image' => $mwPod['spec']['containers'][0]['image'],
+                                'env' => array_merge(
+                                    $mwPod['spec']['containers'][0]['env'],
+                                    [['name' => 'WBS_DOMAIN', 'value' => $this->wikiDomain]]
+                                ),
+                                'command' => [
```
m90, line 69: We could also externalise this script if we're unhappy with having it defined inline? Not sure.

app/Jobs/ProcessMediaWikiJobsJob.php
```diff
@@ -0,0 +1,114 @@
...
+                        ],
+                        'restartPolicy' => 'Never'
+                    ]
+                ]
+            ]
+        ]);
+        $jobName = $kubernetesClient->jobs()->create($jobSpec)['metadata']['name'];
+
+        // TODO: figure out how to use `wait` method instead
+        while (true) {
```
m90, line 95: This is kind of ugly :tm: but it works. There is a `watch` method provided by the Kubernetes client, but it has zero documentation and my PHP knowledge is not good enough to use it in a blocking manner. If someone knows how to use it here, any help is appreciated. This issue https://github.com/maclof/kubernetes-client/issues/58 is the only hint on how it could be used that I found.

config/queue.php
```diff
@@ -63,10 +63,9 @@
             'driver' => 'redis',
             'connection' => 'default',
             'queue' => env('REDIS_QUEUE', 'default'),
-            'retry_after' => 90,
+            'retry_after' => 60 * 60 * 6,
```
m90, line 66: Instead of raising these values globally and still having each item processed on the `default` connection, we could also introduce a dedicated connection for MediaWiki jobs. This would mean we'd need to add an extra deployment for that queue in the `api` Helm chart.

m90 (commented):
Every process spawned by `artisan:queue work` processes jobs in series only. This means that when deploying these changes using the existing configuration, we'd get a hard concurrency limit of one. The prescribed way of having jobs run in parallel is raising the number of replicas for the `queue` pod. I tested this locally and it worked as expected. I just don't really know what a good value for production environments would be. Maybe 8?

----

This PR should probably also still be covered by tests, which I'll be looking into next.

app/Jobs/ProcessMediaWikiJobsJob.php
tarrow: I think yes we should probably want to keep stuff in a separate NS (see following comment too).

I suspect that responsibility for this would *probably* make sense to put in wbaas-deploy in some way rather than here. I would have expected it to be as simple as adding a Namespace yaml to the api chart.

app/Jobs/ProcessMediaWikiJobsJob.php
tarrow: I wonder if we want to; right now this is "fine" but not the easiest to read. I'd probably prefer if we could do something like helmfile does and have templates for these yaml files ? No idea if this is doable though?

app/Jobs/ProcessMediaWikiJobsJob.php
m90: Ah, I somehow thought you couldn't yaml-spec a namespace, but it seems I was wrong: https://kubernetes.io/docs/tasks/administer-cluster/namespaces/ 

In that case, let's put it in the api chart, yes.

tests/Jobs/PollForMediaWikiJobsJobTest.php
```diff
@@ -0,0 +1,51 @@
...
+use App\Wiki;
+use Illuminate\Contracts\Queue\Job;
+use App\Jobs\PollForMediaWikiJobsJob;
+use Illuminate\Foundation\Testing\RefreshDatabase;
+use Illuminate\Support\Facades\Http;
+
+class PollForMediaWikiJobsJobTest extends TestCase
+{
+
+    use RefreshDatabase;
```
m90, line 15: This (stub) test case passes, but makes subsequent tests fail with odd error messages. This makes me think I introduce some statefulness and still need to clean something, but tbh I don't really know what. Any pointers?

app/Jobs/ProcessMediaWikiJobsJob.php
tarrow: What would you think about not running this in a blocking way but simply having the API submit the job and then move on?

We could consider throttling the number of concurrent jobs being flushed through with a solution inside kubernetes like this: https://stackoverflow.com/a/71069007

I think this may be better since a badly behaving job on one wiki is less likely to block up everything and the stability of the api queue instance is less essential to keeping the concurrency controlled (e.g. imagine due to some bug the api queue is regularly restarting we could end up with many jobs being submitted at the same time since really the lock is only held per running process)

app/Jobs/ProcessMediaWikiJobsJob.php
tarrow: I think it would be fine

app/Jobs/PollForMediaWikiJobsJob.php
tarrow: there is a way: we have sufficient credentials in this place to read the db. I personally prefer this though since in my opinion the http is the more stable interface

app/Console/Kernel.php
tarrow: I'm definitely *slightly* worried that this is too slow; maybe it is ok as a first solution though. We could also leave the "running stuff on requests" enabled as well to mitigate this?

I would prefer 1/s up to 1/10s since I think a minute is long enough user would notice the job not having been done (e.g. they start immediately searching for the thing they just did)

app/Jobs/PollForMediaWikiJobsJob.php
tarrow: I believe php doesn't really have this without quite some ugly looking stuff. Laravel provides https://laravel.com/docs/5.5/helpers#method-data-get which I would probably use here

app/Jobs/ProcessMediaWikiJobsJob.php
tarrow: I would guess (crudely) that by including the wiki in the name of the job this would deduplicate (e.g. stop using generate name and use a defined name)

m90: Yeah, I think I like that.

app/Console/Kernel.php
m90: Wouldn't that require its own process as it's supposed to run forever?

app/Jobs/PollForMediaWikiJobsJob.php
```diff
@@ -0,0 +1,46 @@
...
+            return false;
+        }
+
+        $pendingJobs = data_get($response->json(), 'query.statistics.jobs', 0);
+        return $pendingJobs > 0;
+    }
+
+    private function enqueueWiki ( string $wikiDomain ): void
+    {
+        dispatch(new ProcessMediaWikiJobsJob( $wikiDomain ));
```
m90, line 44: @tarrow now that we limit concurrency on Kubernetes level, do you think it still makes sense to fan out to another job here? I'm a bit torn.

Pros I can see:
- cleaner if one job fails, i.e. no need to try/catch here
- reduces execution of this job
- allows for paralellizing if ever needed

Cons:
- Harder to understand
- Performance problems are probably negligible right now, i.e this might just be overengineering

app/Jobs/ProcessMediaWikiJobsJob.php
```diff
@@ -0,0 +1,135 @@
...
+use Illuminate\Support\Facades\Log;
+use Maclof\Kubernetes\Client;
+use Maclof\Kubernetes\Models\Job as KubernetesJob;
+
+class ProcessMediaWikiJobsJob implements ShouldQueue, ShouldBeUnique
+{
+    use InteractsWithQueue, Queueable;
+
+    private string $wikiDomain;
+    private const API_JOB_CONCURRENCY_LIMIT = intval(getenv('API_JOB_CONCURRENCY_LIMIT') || '8', 10);
```
m90, line 18: Is this a good way to read an int value from an env var and also provide a default in case it's not set?

app/Jobs/ProcessMediaWikiJobsJob.php
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    public function uniqueId(): string
+    {
+        return $this->wikiDomain;
+    }
+
+    public function handle (Client $kubernetesClient): void
+    {
+        $kubernetesClient->setNamespace('api-jobs');
```
m90, line 32: I'm not a big fan of the namespace's name. Any ideas?

app/Jobs/ProcessMediaWikiJobsJob.php
m90: Let's maybe keep it like this for now until we figure out the more important stuff.

app/Jobs/ProcessMediaWikiJobsJob.php
```diff
@@ -0,0 +1,135 @@
...
+
+        $numJobs = $kubernetesClient->jobs()->setFieldSelector([
+            'status.phase' => 'Running',
```
m90, line 35: Theoretically, this could also be `Pending`, but I don't have a clear idea how to query for jobs that are either `Running` or `Pending` right now?

m90: It seems it doesn't even work at all like this and we'll have to manually filter out the result sets manually ourselves.

PR #602, opened 2023-05-04
Support two Elasticsearch clusters

AndrewKostka:


rosalieper:
Is this still a WIP?

app/Jobs/CirrusSearch/QueueSearchIndexBatches.php
```diff
@@ -86,12 +97,13 @@ public function handleResponse( string $rawResponse, $error ) : void {
...
         } else {
-
             Log::error(__METHOD__ . ": Job finished but didn't create any batches, something is weird");
-            $this->fail( new \RuntimeException($this->apiModule() . ' call for '.$this->wikiId.' was not successful:' . $rawResponse ) );
+            $this->fail( new \RuntimeException($this->apiModule() . ' call for ' . $this->wikiId() . ' was not successful:' . $rawResponse ) );
         }
+    }
 
+    protected function getQueryParams(): string {
+        return parent::getQueryParams() . '&cluster=' . $this->cluster();
```
m90, line 107: @AndrewKostka I'm currently looking into moving this forward and there's a high-level question I'd like to ask here: what's the reason you decided to have api be in control over which clusters the maintenance scripts are invoked for, when instead this could also be handled by our wrappers in MediaWiki themselves? E.g. by default it seems most CirrusSearch maintenance scripts already work on all known clusters by default.

m90: > AFAIK the CirrusSearch maintenance scripts do not work on all clusters by default.

Interesting. The script level documentation / comments say differently, but it's probably best to be explicit in that case.

PR #603, opened 2023-05-11
When wikis have pending jobs, create a Kubernetes job to work through the items

m90:
Follows up on #601 which my fat fingers accidentally deleted.

----

Ticket https://phabricator.wikimedia.org/T330389

Requires https://github.com/wbstack/charts/pull/114, https://github.com/wmde/wbaas-deploy/pull/885

### DIFF
--- a/app/Jobs/PollForMediaWikiJobsJob.php
+++ b/app/Jobs/PollForMediaWikiJobsJob.php
@@ -4,6 +4,7 @@ namespace App\Jobs;
 
 use App\Wiki;
 use Illuminate\Support\Facades\Http;
+use Illuminate\Support\Facades\Log;
 
 class PollForMediaWikiJobsJob extends Job
 {
@@ -26,12 +27,10 @@ class PollForMediaWikiJobsJob extends Job
         );
 
         if ($response->failed()) {
-            $this->fail(
-                new \RuntimeException(
-                    'Failure polling wiki '.$wikiDomain.' for pending MediaWiki jobs: '.$response->clientError()
-                )
+            $this->job->markAsFailed();
+            Log::error(
+                'Failure polling wiki '.$wikiDomain.' for pending MediaWiki jobs: '.$response->clientError()
             );
-
             return false;
         }
 
--- a/app/Jobs/PollForMediaWikiJobsJob.php
+++ b/app/Jobs/PollForMediaWikiJobsJob.php
@@ -3,6 +3,7 @@
 namespace App\Jobs;
 
 use App\Wiki;
+use Illuminate\Support\Facades\Http;
 
 class PollForMediaWikiJobsJob extends Job
 {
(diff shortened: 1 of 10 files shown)


app/Jobs/PollForMediaWikiJobsJob.php
```diff
@@ -0,0 +1,46 @@
...
+
+    private function hasPendingJobs (string $wikiDomain): bool
+    {
+        $response = Http::withHeaders([
+            'host' => $wikiDomain
+        ])->get(
+            getenv('PLATFORM_MW_BACKEND_HOST').'/w/api.php?action=query&meta=siteinfo&siprop=statistics&format=json'
+        );
+
+        if ($response->failed()) {
```
m90, line 28: Not sure about the error handling here. If one wiki breaks for whatever reason, it would block other wikis. Is there an idiomatic way of returning an error so the iterating loop can collect these, and fail the job after as many wikis as possible have been processed?

m90: E.g. instead of returning `bool` from `hasPendingJobs` I could return `bool|SomeCustomException` and then type switch on the result, but I have no idea if this is considered good practice, or if there are better alternatives?

m90: Maybe it would be enough to log and mark the job as failed here instead?

tests/Routes/Auth/LoginTest.php
```diff
@@ -31,9 +31,10 @@ public function testLoginFail_badPassword()
 
     public function testLoginSuccess()
     {
+        $this->markTestSkipped('fails for unknown reasons');
```
m90, line 34: No idea why this particular test starts failing tbh.

m90: It seems using the `RefreshDatabase` trait in the newly introduced test is now messing the output written by `artisan passport:install`: https://stackoverflow.com/questions/50113508/how-to-test-authentication-via-api-with-laravel-passport/50243105#50243105

app/Jobs/ProcessMediaWikiJobsJob.php
```diff
@@ -0,0 +1,145 @@
...
+{
+    use InteractsWithQueue, Queueable;
+
+    private string $wikiDomain;
+    private int $apiJobConcurrencyLimit;
+
+    public function __construct (string $wikiDomain)
+    {
+        $this->wikiDomain = $wikiDomain;
+        $this->apiJobConcurrencyLimit = getenv('API_JOB_CONCURRENCY_LIMIT')
```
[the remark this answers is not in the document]
m90: `env` is exactly what I've been looking for, thanks for the hint.

app/Jobs/ProcessMediaWikiJobsJob.php
```diff
@@ -0,0 +1,131 @@
...
+
+    public function uniqueId(): string
+    {
+        return $this->wikiDomain;
+    }
+
+    public function handle (Client $kubernetesClient): void
+    {
+        $filterCompletedJobs = function (KubernetesJob $job, int $key): bool {
+            $phase = data_get($job->toArray(), 'status.conditions.0.type', null);
```
m90, line 35: This only seems to work for jobs that are backed by a pod. Jobs that failed creation because they could not be scheduled report `{}` as their status. Not entirely sure right now how to best match these.

deer-wmde (approved):
Can confirm this worked locally for me! details: https://phabricator.wikimedia.org/T330389#8895083

PR #610, opened 2023-06-14
fix(mw-jobs): schedule polling job without overlapping

m90:
Ticket: https://phabricator.wikimedia.org/T339071

I got this working locally (adding a `sleep(100)` to the job itself), but it's very hard to test and not entirely clear to me what happens under the hood. We might need to clear the current queue in production before deploying this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # api
 
+## 8x.12.3 - 13 June 2023
+- Schedule job for polling wikis without overlap
+
 ## 8x.12.2 - 13 June 2023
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # api
 
+## 8x.12.3 - 13 June 2023
```
tarrow, line 3: since merging #611 this is now needing updating if we go in this direction

tarrow (changes requested):
We mustn't get muddled and double use 8x.12.3

PR #613, opened 2023-06-21
feat(wiki): force lower case domain names on creation

m90:
Ticket https://phabricator.wikimedia.org/T339860

I was able to test this locally, both with subdomains as well as custom domain names.

### DIFF
--- a/app/Http/Controllers/WikiController.php
+++ b/app/Http/Controllers/WikiController.php
@@ -32,7 +32,7 @@ class WikiController extends Controller
     public function create(Request $request): \Illuminate\Http\Response
     {
         $user = $request->user();
-        $submittedDomain = $request->input('domain');
+        $submittedDomain = strtolower($request->input('domain'));
         
         $validator = $this->domainValidator->validate( $submittedDomain );
         $isSubdomain = $this->isSubDomain($submittedDomain);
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/WikiController.php
```diff
@@ -32,7 +32,7 @@ public function __construct( DomainValidator $domainValidator )
     public function create(Request $request): \Illuminate\Http\Response
     {
         $user = $request->user();
-        $submittedDomain = $request->input('domain');
+        $submittedDomain = strtolower($request->input('domain'));
```
m90, line 35: Not sure about strings in PHP, does this need to be `strtolower` or `mb_strtolower`?

tests/Routes/Wiki/CreateTest.php
```diff
@@ -51,7 +51,7 @@ public function testWikiCreateDispatchesSomeJobs( $elasticSearchEnabledForNewWik
             'POST',
             $this->route,
             [
-                'domain' => 'derp.com',
+                'domain' => 'dErP.com',
```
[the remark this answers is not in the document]
tarrow: nah, I think leaving it like this is fine

app/Http/Controllers/WikiController.php
tarrow: either would do, I think there is no point in using mb_strlower which I assume is internally more expensive and complex.

This would be necessary if we were to go on to support multibyte chars in this place but since we filter any of those out with the validator I think it doesn't matter (and I don't think we'd ever want to support these in our domain names).

tarrow (approved):
looks good to me. l can confirm this works as I'd expect locally. :)

PR #614, opened 2023-06-21
feat(ci): automatically update image version in wbaas-deploy on tagged release

m90:
Ticket https://phabricator.wikimedia.org/T340039

This works slightly differently than in UI as we still need to push a tag before a release happens.

### DIFF
--- a/.github/workflows/docker.build.yml
+++ b/.github/workflows/docker.build.yml
@@ -55,3 +55,75 @@ jobs:
           push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.docker_meta.outputs.tags }}
           labels: ${{ steps.docker_meta.outputs.labels }}
+
+      -
+        name: Check whether wbaas-deploy should be updated
+        id: update-check
+        run: |
+          if [[ "$GITHUB_REF" =~ ^refs/tags/.+$ ]]; then
+              echo "This is a tagged release, will try to create an update in wbaas-deploy."
+              echo "match=true" >> $GITHUB_OUTPUT
+          fi
+      -
+        name: Check out `wmde/wbaas-deploy` repository in staging child directory
+        if: steps.update-check.outputs.match == 'true'
+        uses: actions/checkout@v3.3.0
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.build.yml
```diff
@@ -54,4 +54,75 @@ jobs:
           cache-to: type=local,mode=max,dest=/tmp/.buildx-cache
           push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.docker_meta.outputs.tags }}
-          labels: ${{ steps.docker_meta.outputs.labels }}
\ No newline at end of file
+          labels: ${{ steps.docker_meta.outputs.labels }}
+
+      -
+        name: Check whether wbaas-deploy should be updated
```
m90, line 60: An alternative to doing this check and then skipping the following steps would be to stop releasing `latest` when pushing to `main` which I think we don't ever use anywhere anyways?

m90:
This needs a powerful person that can set `WBAAS_DEPLOY_TOKEN` in the repository secrets before merging.

m90:
A token is now added to this repo.

.github/workflows/docker.build.yml
tarrow: yeah, we don't use it anyway either AFAICT; I believe we should therefore feel free to remove it.

tarrow: ah, yeah. I believe we definitely want to check this builds on each PR (otherwise there is a high chance of us merging non-building code).

One alternative if you wanted to refactor more heavily would be to:
- stick these following steps in their own job.
- Make this job dependent on the build job running (https://docs.github.com/en/actions/using-jobs/using-jobs-in-a-workflow#example-requiring-successful-dependent-jobs)
-  using a condition on the job to ensure it only runs on tags that exist on `main` (https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution)

tarrow (approved):
Would be happy to see merged as is or to review again if you prefer to think about some alterations after reading the one comment thread

PR #619, opened 2023-07-10
feat(signup): throttle user account creation

m90:
Ticket https://phabricator.wikimedia.org/T335963

Requires https://github.com/wmde/wbaas-deploy/pull/983 https://github.com/wbstack/charts/pull/121 https://github.com/wbstack/ui/pull/700

### DIFF
--- a/app/Http/Middleware/ThrottleSignup.php
+++ b/app/Http/Middleware/ThrottleSignup.php
@@ -22,7 +22,7 @@ class ThrottleSignup
             Log::warning("WARN_SIGNUP_THROTTLED: Given limit of '$limit' in range '$range' was exceeded, attempted account creation was blocked.");
             return response()->json(
                 ["errors" => "Due to high load, we're currently not able to create an account for you. Please try again tomorrow or reach out through our contact page."],
-                429
+                503
             );
         }
 
--- a/app/Http/Middleware/ThrottleSignup.php
+++ b/app/Http/Middleware/ThrottleSignup.php
@@ -6,6 +6,7 @@ use Closure;
 use Carbon\Carbon;
 use App\User;
 use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Log;
  
 class ThrottleSignup
 {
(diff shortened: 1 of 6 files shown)


app/Http/Middleware/ThrottleSignup.php
```diff
@@ -0,0 +1,31 @@
...
+            return $next($request);
+        }
+
+        $lookback = Carbon::now()->sub(new \DateInterval($range));
+        $recentSignups = User::where('created_at', '>=', $lookback)->count();
+        if ($recentSignups >= intval($limit)) {
+            Log::warning("WARN_SIGNUP_THROTTLED: Given limit of '$limit' in range '$range' was exceeded, attempted account creation was blocked.");
+            return response()->json(
+                ["errors" => "Due to high load, we're currently not able to create an account for you. Please try again tomorrow or reach out through our contact page."],
+                429
```
m90, line 25: Not sure if a 429 is super correct here, but everything else 4xx fits even worse https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses

m90: Maybe 503 would be a better choice as it's not really the client's fault.

config/wbstack.php
```diff
@@ -19,4 +19,7 @@
     'elasticsearch_host' => env('ELASTICSEARCH_HOST', false),
     'elasticsearch_enabled_by_default' => env('WBSTACK_ELASTICSEARCH_ENABLED_BY_DEFAULT', false),
 
-];
\ No newline at end of file
+    'signup_throttling_limit' => env('SIGNUP_THROTTLING_LIMIT', ''),
+    'signup_throttling_range' => env('SIGNUP_THROTTLING_RANGE', ''),
```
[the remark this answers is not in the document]
m90: That's totally right, fixed it.

app/Http/Middleware/ThrottleSignup.php
m90: Yeah, I'll sneak this in still before merging.

PR #621, opened 2023-07-20
feat(queue): Add failed job backoff configuration

deer-wmde:
Currently failed jobs will be retried 5 times[1]. This adds job backoff configuration as suggested [here](https://phabricator.wikimedia.org/T325894#9023530), so that failed retries will happen with incremental delays:
- 10 seconds
- 1 minute
- 5 minutes
- 15 minutes
- 30 minutes

[1] https://github.com/wbstack/api/blob/main/start.sh#L20

Draft status because I couldn't make this work on my local cluster yet.

### DIFF
--- a/app/Jobs/Job.php
+++ b/app/Jobs/Job.php
@@ -6,6 +6,7 @@ use Illuminate\Bus\Queueable;
 use Illuminate\Contracts\Queue\ShouldQueue;
 use Illuminate\Queue\InteractsWithQueue;
 use Illuminate\Queue\SerializesModels;
+use Illuminate\Support\Facades\Config;
 
 abstract class Job implements ShouldQueue
 {
@@ -20,4 +21,9 @@ abstract class Job implements ShouldQueue
     |
     */
     use InteractsWithQueue, Queueable, SerializesModels;
+
+    public function backoff(): array
+    {
+        return Config::get('queue.backoff');
+    }
 }
--- a/config/queue.php
+++ b/config/queue.php
@@ -84,5 +84,5 @@ return [
         'table' => 'failed_jobs',
     ],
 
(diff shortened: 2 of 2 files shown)


deer-wmde:
> When trying to test this, do you make jobs fail by calling `$job->fail()` or by throwing an exception? Not sure if it makes a difference, but maybe it's a first idea where to start debugging.

I tried testing without changing code, I tried to create wikis while no mediawiki pods are around (so I guess this is equivalent to the exception throwing). Will try those two ways explicitly, I didn't know about the $job->fail method

app/Jobs/Job.php
```diff
@@ -20,4 +21,9 @@ abstract class Job implements ShouldQueue
     |
     */
     use InteractsWithQueue, Queueable, SerializesModels;
+
+    public function backoff(): array
+    {
+        return Config::get('queue.backoff');
```
m90, line 27: Do you have any idea what happens in case this is an empty array? Will it just retry immediately?

config/queue.php
```diff
@@ -84,4 +84,5 @@
         'table' => 'failed_jobs',
     ],
 
+    'backoff' => explode(',', env('QUEUE_BACKOFF', '10,60,300,900,1500')), // 10s, 1m, 5m, 15m, 30m
```
m90, line 87: If someone would set QUEUE_BACKOFF to the empty string, would this use the default? I'm wondering if we should have a null default so that we can turn this behavior on and off by means of configuration.

app/Jobs/Job.php
deer-wmde: Just tried this, yes

config/queue.php
deer-wmde: Good thinking, I removed the values

m90 (approved):

PR #622, opened 2023-07-20
feat(queue): `report()` on failed jobs

deer-wmde:
This adds calling the Laravel `report()` helper function [1] as error handling for failed queue'd jobs as per https://phabricator.wikimedia.org/T325894#9023483.

Note: `$event->job->getName();` was changed to `data_get($event->job->payload(), 'data.commandName');`, as the former would always return a constant string for the class `CallQueuedHandler`. See https://github.com/illuminate/queue/blob/8.x/Queue.php#L143

[1] https://laravel.com/docs/8.x/helpers#method-report

### DIFF
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -3,6 +3,8 @@
 namespace App\Providers;
 
 use Illuminate\Support\ServiceProvider;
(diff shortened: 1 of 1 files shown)


m90 (approved):
Let's try this I guess :shipit:

PR #623, opened 2023-07-24
refactor(MediawikiInitJob): remove fail() calls

deer-wmde:
https://phabricator.wikimedia.org/T325894

Our Laravel Jobs won't get retried if we mark them as `failed` via `$this->fail()` if erros occur during the jobs processing.

This PR removes unwraps those calls and throws the Exceptions for the `MediawikiInit` Job. This way, the job gets retried in case it fails.

For more effectiveness, this should not be released before https://github.com/wbstack/api/pull/621

### DIFF
--- a/tests/Jobs/MediawikiInitTest.php
+++ b/tests/Jobs/MediawikiInitTest.php
@@ -82,11 +82,11 @@ class MediawikiInitTest extends TestCase
 
         $mockJob = $this->createMock(Job::class);
 
-        $this->expectException(\RuntimeException::class);
-        $this->expectExceptionMessage($mockExceptionMessage);
-
         $job = new MediawikiInit( $this->wikiDomain, $this->username, $this->email );
         $job->setJob($mockJob);
+
+        $this->expectException(\RuntimeException::class);
+        $this->expectExceptionMessage($mockExceptionMessage);
         $job->handle($request);
     }
 }
--- a/tests/Jobs/MediawikiInitTest.php
+++ b/tests/Jobs/MediawikiInitTest.php
@@ -76,8 +76,9 @@ class MediawikiInitTest extends TestCase
     public function testFatalErrorIsHandled()
     {
         $mockResponse = 'oh no';
-        $mockExceptionMessage = 'wbstackInit call for some.domain.com. No wbstackInit key in response: ' . $mockResponse;
         $request = $this->getMockRequest($mockResponse);
+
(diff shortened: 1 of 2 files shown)


tests/Jobs/MediawikiInitTest.php
```diff
@@ -76,16 +76,22 @@ public function testSuccess()
...
 
         $mockJob = $this->createMock(Job::class);
-        $mockJob->expects($this->once())
-                ->method('fail')
-                ->with(new \RuntimeException('wbstackInit call for some.domain.com. No wbstackInit key in response: ' . $mockResponse ));
 
-        $job = new MediawikiInit( $this->wikiDomain, $this->username, $this->email );
-        $job->setJob($mockJob);
-        $job->handle($request);
+        $caughtException = null;
```
deer-wmde, line 85: This test works as is, but feels not right to me. Earlier I tried something like `$mockJob->method('handle')->willThrowException ...` but this didn't work out for me: 
```
1) Tests\Jobs\MediawikiInitTest::testFatalErrorIsHandled
PHPUnit\Framework\MockObject\MethodCannotBeConfiguredException: Trying to configure method "handle" which cannot be configured because it does not exist, has not been specified, is final, or is static
```

m90: Without having tried it, I think you can use `expectException`: https://phpunit.de/manual/6.5/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.exceptions.examples.ExceptionTest.php

tests/Jobs/MediawikiInitTest.php
```diff
@@ -76,16 +76,17 @@ public function testSuccess()
     public function testFatalErrorIsHandled()
     {
         $mockResponse = 'oh no';
-
         $request = $this->getMockRequest($mockResponse);
 
+        $mockExceptionMessage = 'wbstackInit call for some.domain.com. No wbstackInit key in response: ' . $mockResponse;
```
m90, line 81: Nit: this is not a mock, but an expectation. Maybe `expectedExceptionMessage` would be more fitting?

m90 (approved):
Left one tiny remark, everything else looks ✨ it's really great you found out about this entire condition.

tests/Jobs/MediawikiInitTest.php
deer-wmde: yes! agreed

deer-wmde:
> Left one tiny remark, everything else looks ✨ it's really great you found out about this entire condition.

Yes! Thanks for your help, you gave me just the right hints the whole time :)

PR #626, opened 2023-08-08
fix: replace recaptcha package

deer-wmde:
https://phabricator.wikimedia.org/T335769#9067453

This PR fixes ReCaptcha validation by replacing the package `albertcht/invisible-recaptcha` with the official `google/recaptcha` package, used via a Validation Rule class and a Service Provider.

Using the dependency injection allowed us to replace the (broken) [change in application path for testing](https://github.com/wbstack/api/blob/62f47d57eafb0e5feb0ba99830ec859496716f7c/app/Http/Controllers/Auth/RegisterController.php#L74)  with mocking the ReCaptcha library. 

Other changes:
- read log level from env var, default to `info` (currently it's `debug` everywhere)
- rename ReCaptcha env vars
- set app environment to `testing` for CI
- some cleanup here and there

I'm not quite happy with the mocking copypasta in the tests, but after lots of tries to reuse the code I think a working way would be to add a Facade for the ReCaptchaValidation Service and implement a Facade and Fake class (see [NotificationFake](https://github.com/laravel/framework/blob/8.x/src/Illuminate/Support/Testing/Fakes/NotificationFake.php) for example) but that sounds a bit much too me right now


app/Http/Controllers/ContactController.php
```diff
@@ -50,10 +60,6 @@ public function sendMessage(Request $request): \Illuminate\Http\JsonResponse
      */
     protected function validator(array $data): \Illuminate\Validation\Validator
     {
-        if (! isset($data['contactDetails'])) {
-            $data['contactDetails'] = ''; // could we skip this using some feature of the validator
```
tarrow, line 54: Maybe I missed it but how did we remove the need for this workaround?

app/Rules/ReCaptchaValidation.php
```diff
@@ -0,0 +1,125 @@
...
+        $this->minScore = $minScore;
+    }
+
+    /**
+     * Comparison against expected hostname.
+     * 
+     * @param  string $hostname
+     * @return bool  
+     */
+    public function verifyHostname($hostname)
```
tarrow, line 38: I think this could probably be private?

app/Rules/ReCaptchaValidation.php
```diff
@@ -0,0 +1,125 @@
...
+        return true;
+    }
+
+    /**
+     * Verifies the ReCaptcha Request with the official ReCaptcha service library
+     * 
+     * @param  string $secretKey
+     * @return \ReCaptcha\Response 
+     */
+    public function verify($token)
```
tarrow, line 60: I think this could be private?

tests/Routes/User/RegisterTest.php
```diff
@@ -19,20 +20,44 @@ class RegisterTest extends TestCase
 
     // TODO test password length when not deving
 
+    public function mockReCaptchaValidation($fakeResponse=[])
```
tarrow, line 23: I know you didn't like having to copypasta this from the PR description. I wondered if an alternative that would be a little smaller would be to have a mock rule that only had the `passes` method mocked rather than the extra complexity of mocking the `verify` method?

I think that would make sense since each of these tests really doesn't want to know about the internal of the Rule (incidentally making the `verifiy` method private would prevent this systematically)

In this case you could perhaps almost inline it in any given test with *something* like
```
$mockValidation = $this->createMock(ReCaptchaValidation::class);
$mockValidation->method('passes')->willReturn(true) // or false
$this->app->instance(ReCaptchaValidation::class, $mockValidation);
```

(above was slapped together sans IDE so might need a little tweaking)

app/Rules/ReCaptchaValidation.php
```diff
@@ -0,0 +1,125 @@
...
+    {
+        $recaptchaResponse = new \ReCaptcha\Response(false);
+
+        try {
+            $recaptcha = new \ReCaptcha\ReCaptcha(
```
tarrow, line 65: I wonder if this dependency would be better injected into the constructor? This would better enable us to test it?

I think that's probably why writing the tests for this class feels tricky and perhaps why it felt necessary to expose `verify` as a public method for tinkering in the tests.

tests/Validation/ReCaptchaValidationTest.php
```diff
@@ -0,0 +1,106 @@
...
+namespace Tests\Feature;
+
+use Illuminate\Foundation\Testing\RefreshDatabase;
+use Illuminate\Foundation\Testing\WithFaker;
+use Tests\TestCase;
+use App\Rules\ReCaptchaValidation;
+
+class ReCaptchaValidationTest extends TestCase
+{
+    public function buildMockedReCaptchaRule($fakeResponse)
```
tarrow, line 12: Given we are supposed to be testing the RecaptchaRule its unfortunate that we need to mock parts of it to test it.

I was thinking about that we could inject the `\ReCaptcha\ReCaptcha` into the rule (see above comment on implementation of class). This would then enable us to mock just this class from the library to test similar functionality but also the logic of the `verify` method.

tarrow:
This is really excellent! I love the use of the "official" upstream library rather than the 3rd party one.

Code was very readable and understandable; enjoyed reviewing this.

Hopefully the suggestion to make the controller tests know even less about the internal of the rule might relieve your copy pasta worry.

app/Http/Controllers/ContactController.php
deer-wmde: I could've sworn I seen this work without it earlier, but I just added a test that proved me otherwise.

app/Rules/ReCaptchaValidation.php
deer-wmde: I think so, yep

app/Rules/ReCaptchaValidation.php
deer-wmde: Ah yeah that makes sense. All this code went through several iterations and at one point this class was instantiated manually in the Controllers and I wanted to minimize that setup. But now it can happen in the ServiceProvider :+1: will change it

tests/Routes/User/RegisterTest.php
deer-wmde: ooh that's nice!

tests/Routes/Contact/SendMessageTest.php
```diff
@@ -39,16 +37,34 @@ class SendMessageTest extends TestCase
         'other',
     ];
 
+    public function mockReCaptchaValidation($passes=true)
+    {
+        // replace injected ReCaptchaValidation class with mock (ContactController::$recaptchaValidation)
+        $mockRuleBuilder = $this->getMockBuilder(ReCaptchaValidation::class);
```
tarrow, line 43: As we just said in person this could be replaced by a $createMock

tests/Routes/User/RegisterTest.php
```diff
@@ -19,20 +20,34 @@ class RegisterTest extends TestCase
 
     // TODO test password length when not deving
 
+    public function mockReCaptchaValidation($passes=true)
+    {
+        // replace injected ReCaptchaValidation class with mock (ContactController::$recaptchaValidation)
+        $mockRuleBuilder = $this->getMockBuilder(ReCaptchaValidation::class);
```
tarrow, line 26: same as above :)

tests/Routes/User/RegisterTest.php
```diff
@@ -19,20 +20,30 @@ class RegisterTest extends TestCase
 
     // TODO test password length when not deving
 
+    public function mockReCaptchaValidation($passes=true)
```
tarrow, line 23: such nitpick: can be private I think

tests/Routes/Contact/SendMessageTest.php
```diff
@@ -39,16 +37,30 @@ class SendMessageTest extends TestCase
         'other',
     ];
 
+    public function mockReCaptchaValidation($passes=true)
```
tarrow, line 40: such nitpick: can be private

app/Rules/ReCaptchaValidation.php
```diff
@@ -0,0 +1,129 @@
...
+        $this->appUrl = $appUrl;
+    }
+
+    /**
+     * Comparison against expected hostname.
+     * 
+     * @param  string $hostname
+     * @return bool  
+     */
+    private function verifyHostname($hostname)
```
tarrow, line 39: I just realised here: for example you could have type hinted using more than just docstring (e.g. actually type bot the parameters and return types; no need to fix but maybe we want to start linting for this for consistency

tarrow (approved):
tiny nitpicks; feel free to fix or not and merge

PR #627, opened 2023-08-10
chore(jobs): stats collection at 7AM does not conflict with backup schedule

m90:
Ticket https://phabricator.wikimedia.org/T317980

### DIFF
--- a/app/Console/Kernel.php
+++ b/app/Console/Kernel.php
@@ -11,9 +11,6 @@ use App\Jobs\SandboxCleanupJob;
 use App\Jobs\PollForMediaWikiJobsJob;
 use Illuminate\Console\Scheduling\Schedule;
 use Illuminate\Foundation\Console\Kernel as ConsoleKernel;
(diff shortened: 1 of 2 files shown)


app/Console/Kernel.php
```diff
@@ -43,7 +43,7 @@ protected function schedule(Schedule $schedule)
         $schedule->job(new SandboxCleanupJob)->everyFifteenMinutes();
 
         // Schedule site stat updates for each wiki and platform-summary
-        $schedule->command('schedule:stats')->daily();
+        $schedule->command('schedule:stats')->dailyAt('7:00');
```
m90, line 46: This triggers jobs for both platform summary and per site stats: https://github.com/wbstack/api/blob/e9546668cd92f937bebf0885ab700ac6cfda9f78/app/Console/Commands/ScheduleStatsUpdates.php#L54-L57

rosalieper (approved):

PR #630, opened 2023-08-14
Respond with appropriate status codes on failing wiki lookup in backend

m90:
Ticket: https://phabricator.wikimedia.org/T343743

### DIFF
--- a/app/Http/Controllers/Backend/WikiController.php
+++ b/app/Http/Controllers/Backend/WikiController.php
@@ -18,7 +18,7 @@ class WikiController extends Controller
         try {
             if ($domain === 'localhost' || $domain === 'mediawiki') {
                 // If just using localhost then just get the first undeleted wiki
-                $result = Wiki::with(self::$with)->limit(1)->first();
+                $result = Wiki::with(self::$with)->first();
             } else {
                 // TODO don't select the timestamps and redundant info for the settings?
                 $result = Wiki::where('domain', $domain)->with(self::$with)->first();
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/Backend/WikiController.php
```diff
@@ -10,22 +10,32 @@ class WikiController extends Controller
 {
     private static $with = ['wikiDb', 'wikiQueryserviceNamespace', 'settings'];
 
-    public function getWikiForDomain(Request $request): \Illuminate\Http\Response
+    public function getWikiForDomain(Request $request): \Illuminate\Http\JsonResponse
```
m90, line 13: I assume this change is ok as we really expect JSON here https://github.com/wbstack/mediawiki/blob/b864439a68bf29f9d81a124dc83a019ef3e92673/dist-persist/wbstack/src/Info/GlobalSet.php#L128

app/Http/Controllers/Backend/WikiController.php
```diff
@@ -10,22 +10,27 @@ class WikiController extends Controller
...
-        } else {
-            // TODO don't select the timestamps and redundant info for the settings?
-            $result = Wiki::where('domain', $domain)->with(self::$with)->first();
+        try {
+            if ($domain === 'localhost' || $domain === 'mediawiki') {
+                // If just using localhost then just get the first undeleted wiki
+                $result = Wiki::with(self::$with)->first();
+            } else {
+                // TODO don't select the timestamps and redundant info for the settings?
+                $result = Wiki::where('domain', $domain)->with(self::$with)->first();
```
m90, line 24: The ticket makes it sound as if there could be multiple wikis per domain, but the column has a uniqueness constraint on schema level so there will only ever be one result if any.

deer-wmde (approved):
excellent work!

PR #631, opened 2023-08-15
feat(console): add Command for changing user mail

deer-wmde:
I wanted to explore the Laravel Console Commands a bit and changing user mails seemed like a nice test subject for reducing points of failures.

So far the process for doing this is
- manually editing the user field on *production*
  - via a `artisan tinker` session
  - or changing it directly in SQL

The idea is to call the command like
```
php artisan wbs-user:change-mail --from='oldmail@example.com' --to='newmail@example.com'
```

After the change is made to the user entity, a verification mail is sent, because the users mail also gets unverified in the process (it's a new address after all; but we should ask Product and UX if this is actually preferred)

### DIFF
--- a/app/Console/Commands/User/ChangeEmail.php
+++ b/app/Console/Commands/User/ChangeEmail.php
@@ -14,8 +14,8 @@ class ChangeEmail extends Command
 
     public function handle(): int
     {
-        $emailOld = $this->argument('--from');
-        $emailNew = $this->argument('--to');
+        $emailOld = $this->option('from');
+        $emailNew = $this->option('to');
 
         $user = User::whereEmail($emailOld)->first();
 
--- a/app/Console/Commands/User/ChangeEmail.php
+++ b/app/Console/Commands/User/ChangeEmail.php
@@ -21,7 +21,7 @@ class ChangeEmail extends Command
             if ($user) {
                 $this->line("Found a user for '$emailOld'");
             } else {
-                $this->warn("Did not find a user for '$emailOld' Please try again.");
+                $this->warn("Did not find a user for '$emailOld'. Please try again.");
             }
         } while(!$user);
 
--- a/app/Console/Commands/User/ChangeEmail.php
+++ b/app/Console/Commands/User/ChangeEmail.php
@@ -8,34 +8,25 @@ use App\Jobs\UserVerificationCreateTokenAndSendJob;
 
 class ChangeEmail extends Command
 {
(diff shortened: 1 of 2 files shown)


app/Console/Commands/User/ChangeEmail.php
```diff
@@ -0,0 +1,56 @@
...
+        do {
+            $emailOld = $this->ask('Current user address');
+            $user = User::whereEmail($emailOld)->first();
+
+            if ($user) {
+                $this->line("Found a user for '$emailOld'");
+            } else {
+                $this->warn("Did not find a user for '$emailOld'. Please try again.");
+            }
+        } while(!$user);
```
m90, line 26: Can one Ctrl+C out of such a loop as expected?

m90 (commented):
How hard would it be to make the command available in both an interactive version (as it is right now), as well as enabling passing everything it needs to know using command line args. I do like the current UX, but it cannot be used in scripts et al, so maybe having both would be nice?

app/Console/Commands/User/ChangeEmail.php
deer-wmde: yes, you can abort it safely anytime before confirming the confirmation prompt

deer-wmde:
> How hard would it be to make the command available in both an interactive version (as it is right now), as well as enabling passing everything it needs to know using command line args. I do like the current UX, but it cannot be used in scripts et al, so maybe having both would be nice?

I also thought about this and I think it's not too much effort to add this, so I will look into it

The idea behind using the prompts was I thought it could be nice to not have to remember how it works (because we use this once every other month so far). But I guess that is not much worth if its not a pattern more widely used, so a part of me is thinking if it's even worth having the prompts in the first place

m90:
> so a part of me is thinking if it's even worth having the prompts in the first place

I'd say non-interactive should be the default, but if we can have both, it's definitely worth it.

deer-wmde:
changed the behavior to non-interactive - I think this is ready for review now

tests/Commands/User/ChangeEmailTest.php
```diff
@@ -0,0 +1,94 @@
+<?php
```
tarrow, line 1: Nice set of test cases covering all that I'm interesting in being tested

tarrow (approved):
Looks good to me; not tested locally but the testcase and the support from the framework leave me pretty confident. Looks so much nicer and safer than hitting tinker in prod

PR #634, opened 2023-08-22
feat(jobs): add job that populates wiki site stats

m90:
Ticket https://phabricator.wikimedia.org/T344583

### DIFF
--- a/app/Console/Kernel.php
+++ b/app/Console/Kernel.php
@@ -9,6 +9,7 @@ use App\Jobs\PruneEventPageUpdatesTable;
 use App\Jobs\PruneQueryserviceBatchesTable;
 use App\Jobs\SandboxCleanupJob;
 use App\Jobs\PollForMediaWikiJobsJob;
+use App\Jobs\UpdateWikiSiteStatsJob;
 use Illuminate\Console\Scheduling\Schedule;
 use Illuminate\Foundation\Console\Kernel as ConsoleKernel;
 
@@ -43,6 +44,8 @@ class Kernel extends ConsoleKernel
         $schedule->command('schedule:stats')->dailyAt('7:00');
 
         $schedule->job(new PollForMediaWikiJobsJob)->everyFifteenMinutes();
+
+        $schedule->job(new UpdateWikiSiteStatsJob)->dailyAt('19:00');
     }
 
     /**
--- a/app/Jobs/UpdateWikiSiteStatsJob.php
+++ b/app/Jobs/UpdateWikiSiteStatsJob.php
@@ -1,0 +1,49 @@
+<?php
+
+namespace App\Jobs;
+
+use App\Wiki;
+use App\WikiSiteStats;
+use Illuminate\Support\Facades\Http;
+use Illuminate\Support\Facades\Log;
+use Illuminate\Contracts\Queue\ShouldBeUnique;
+
+class UpdateWikiSiteStatsJob extends Job implements ShouldBeUnique
+{
+    public $timeout = 3600;
+    public function handle (): void
+    {
+        $allWikis = Wiki::all();
+        foreach ($allWikis as $wiki) {
+            $this->updateSiteStats($wiki);
+        }
+    }
+
+    private function updateSiteStats (Wiki $wiki): void
+    {
+        $response = Http::withHeaders([
+            'host' => $wiki->getAttribute('domain')
+        ])->get(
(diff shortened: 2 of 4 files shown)


app/Jobs/UpdateWikiSiteStatsJob.php
```diff
@@ -0,0 +1,49 @@
...
+            Log::error(
+                'Failure polling wiki '.$wiki->getAttribute('domain').' for sitestats: '.$response->clientError()
+            );
+            return;
+        }
+
+        $responseBody = $response->json();
+        $update = [];
+        foreach (WikiSiteStats::FIELDS as $field) {
+            $value = data_get($responseBody, 'query.statistics.'.$field, null);
```
m90, line 41: I cannot recall what we discussed in the breakdown session: this means if a key is missing, we'll use stale data instead of 0. If we want 0, we can still change it.

app/Console/Kernel.php
```diff
@@ -43,6 +44,8 @@ protected function schedule(Schedule $schedule)
         $schedule->command('schedule:stats')->dailyAt('7:00');
 
         $schedule->job(new PollForMediaWikiJobsJob)->everyFifteenMinutes();
+
+        $schedule->job(new UpdateWikiSiteStatsJob)->dailyAt('19:00');
```
m90, line 48: This is totally arbitrary.

dati18 (changes requested):

app/Jobs/UpdateWikiSiteStatsJob.php
```diff
@@ -0,0 +1,52 @@
...
+
+use App\Wiki;
+use App\WikiSiteStats;
+use Illuminate\Support\Facades\Http;
+use Illuminate\Support\Facades\Log;
+use Illuminate\Contracts\Queue\ShouldBeUnique;
+
+class UpdateWikiSiteStatsJob extends Job implements ShouldBeUnique
+{
+    public $timeout = 3600;
```
dati18, line 13: I don't know if 1hr is enough to run the job in prod, some wikis in prod are huge. No need to change it now but could be something we need to observe

app/Console/Kernel.php
m90: We just recently moved another job _off_ the midnight schedule to reduce load, so I would think it's wise not to run this at midnight as well.

https://phabricator.wikimedia.org/T317980

I'd suggest keeping this at 19:00. Considering we have users around the globe, it's also not really possible to tell when people are sleeping or not.

app/Jobs/UpdateWikiSiteStatsJob.php
m90: We just poll the stats, so this is relatively cheap. If we ever hit the hour, we're doing something wrong.

app/Jobs/UpdateWikiSiteStatsJob.php
m90: Could you elaborate on what you mean exactly? Would you change this behavior?

m90: For the record, we can't `null` the value itself as it's a numeric non-nullable field.

dati18: sorry I for the misunderstanding, I meant `stale data`. I don't know why I said `null`

m90: Ok, so this is currently using stale data already. Do we need to change anything?

app/Console/Kernel.php
dati18: I am not aware how many jobs we run during the night. But if you said so, let's just keep 19:00

m90: The Logical SQL backup (which is not a Laravel) job is running at Midnight and is pretty expensive, so it's best not to schedule anything else during that period https://github.com/wmde/wbaas-deploy/blob/b98f2fcf235a7e56ac1a8b889fc7a13f04be6df8/k8s/helmfile/env/production/wbaas-backup.values.yaml.gotmpl#L6-L7

app/Jobs/UpdateWikiSiteStatsJob.php
dati18: nope, I prefer using stale date instead of 0 because 0 can cause confusion

m90:
@dati18 Anything left to do here?

dati18 (approved):

dati18:
> @dati18 Anything left to do here?

I approved, roll it out :)

PR #635, opened 2023-08-22
feat(api): add public endpoint for querying wikis

m90:
Ticket https://phabricator.wikimedia.org/T344581


app/Http/Controllers/PublicWikiController.php
```diff
@@ -0,0 +1,63 @@
...
+    {
+        return response()->json(['message' => 'Method not allowed'], 405);
+    }
+
+    /**
+     * Display the specified resource.
+     */
+    public function show($id)
+    {
+        return new PublicWikiResource(Wiki::findOrFail($id));
```
m90, line 45: This allows clients to look up wikis by their integer id. Instead, we could also look them up by domain, not sure if I have a preference.

app/Wiki.php
```diff
@@ -57,6 +57,8 @@ class Wiki extends Model
         'deleted_at',
     ];
 
+    protected $perPage = 20;
```
m90, line 60: Not sure if this override is even needed. It defaults to 15 and I mainly put this value here for testing purposes. In the real world, I would expect the UI to always pass a `per_page` query param to tell the endpoint about its preferences.

app/Http/Controllers/PublicWikiController.php
```diff
@@ -0,0 +1,68 @@
...
+
+        $query = Wiki::query();
+
+        $isFeatured = $request->query('is_featured', null);
+        if ($isFeatured !== null) {
+            $query = $query->where(['is_featured' => boolval($isFeatured)]);
+        }
+
+        $isActive = $request->query('is_active', null);
+        if ($isActive !== null && boolval($isActive)) {
```
m90, line 30: This seems terrible, but my PHP is not good enough to come up with something better. Is there any more idiomatic way to do this?

routes/api.php
```diff
@@ -22,6 +22,8 @@
     $router->post('user/resetPassword', ['uses' => 'Auth\ResetPasswordController@reset']);
     $router->post('contact/sendMessage', ['uses' => 'ContactController@sendMessage']);
 
+    $router->apiResource('wiki', 'PublicWikiController');
```
m90, line 25: Not sure if this needs stricter throttling than the default one? Current settings allows for 45 request to be made in one minute per IP. I am not sure however if we actually see the forwarded IP or if this is effectively a global setting, throttling the "ingress client".

app/Http/Controllers/PublicWikiController.php
```diff
@@ -0,0 +1,75 @@
...
+     */
+    public function index(Request $request)
+    {
+        $request->validate([
+            'sort' => 'in:sitename,pages',
+            'direction' => 'in:desc,asc',
+            'is_featured' => 'boolean',
+            'is_active' => 'boolean',
+            'per_page' => 'numeric',
+            'page' => 'numeric'
```
m90, line 27: This is somehow magically picked up by Laravel itself, we never pass it on or look at it. Not sure if we need to validate it, but I thought it might be good to put it here so all possible query params are documented here.

app/Http/Controllers/PublicWikiController.php
```diff
@@ -0,0 +1,75 @@
...
+        $query = Wiki::query();
+
+        if (array_key_exists('is_featured', $params)) {
+            $query = $query->where([
+                'is_featured' => boolval($params['is_featured'])
+            ]);
+        }
+
+        if (array_key_exists('is_active', $params) && $params['is_active']) {
+            $query = $query->whereRelation('wikiSiteStats', 'pages', '>', 0);
```
m90, line 40: The ticket currently say `> 1`, I need to double check if this is correct or needs to be changed.

https://phabricator.wikimedia.org/T339334

m90: The ticket is correct. I fixed this to be `> 1`

dati18 (approved):
Everything is great in my eyes. Didn't catch anything off

PR #636, opened 2023-08-23
fix: replace docker compose with docker-compose

dati18:


m90:
This is intentional. `compose` is now a Docker plugin instead of a standalone tool (which was called `docker-compose`): https://docs.docker.com/compose/install/linux/

On my setup I have the compose plugin installed and defined a `alias docker-compose="docker compose"` so scripts that reference the old style compose will still work.

dati18:
> This is intentional. `compose` is now a Docker plugin instead of a standalone tool (which was called `docker-compose`): https://docs.docker.com/compose/install/linux/
> 
> On my setup I have the compose plugin installed and defined a `alias docker-compose="docker compose"` so scripts that reference the old style compose will still work.

I think we should move to `docker-compose`. People can use Makefile without being confused and doing and extra step of creating an alias

m90:
> I think we should move to docker-compose. People can use Makefile without being confused and doing and extra step of creating an alias

The alias is just for backwards compatibility. You don't need it if you have the newest version of compose installed.

dati18:
ok, good point, I will close the PR ;)

deer-wmde:
This blog article from docker may give some more insight, what I got from it is that `docker-compose` is considered deprecated https://www.docker.com/blog/announcing-compose-v2-general-availability/

PR #637, opened 2023-08-23
feat(http): allow trusted proxies to be defined

m90:
After deploying #635 to staging I noticed links in pagination responses would be using the `http` protocol, when the application itself is served using `https`. This seems to be due to the fact that as configured, Laravel does not trust any proxies at all and will think the client is the ingress, which rewrites to `http`: https://github.com/wbstack/charts/blob/af3206a791b345a70dd8b1670e20ff3d4e30d57a/charts/api/templates/ingress.yaml#L26-L37

This PR introduces the possibility of configuring trusted proxies as documented here: https://laravel.com/docs/8.x/requests#configuring-trusted-proxies

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # api
 
+## 8x.21.0 - 24 August 2023
+- Allow for trusted proxies to be configured
+
 ## 8x.20.0 - 23 August 2023
 - Add public /wiki endpoint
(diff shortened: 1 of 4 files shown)


m90:
Chart PR here https://github.com/wbstack/charts/pull/128

deer-wmde (approved):

PR #644, opened 2023-09-13
fix: injected k8s client should not be a singleton

m90:
Ticket https://phabricator.wikimedia.org/T345605

Having the client injected as a singleton means that users would need to "clean up" after their usage. As the client is stateful and remembers for example the targeted namespace, this can lead to conflicts between users in case the instance would be reused. Making this a new instance each time should be more predictable.


rosalieper (approved):
thank you

PR #646, opened 2023-09-18
Increase EventPageUpdate pruning rate

tarrow:
This commit multiplies by 5 the rate at which we prune events.

Given our long term average prune rate and remaining EventPageUpdates this should be fast enough to keep us pruning to the 100k level we're targeting


m90 (approved):
Sounds sensible to me, although we might want to make this configurable at some point instead of hardcoding the value maybe?

PR #647, opened 2023-09-18
Increase QsBatch prune rate

tarrow:
This commit increases the rate we prune batches older than 1 month by 5 times.

I hope but have not calculated that this is sufficient to keep us pruning at the rate we are generating Batches.


m90 (approved):
Same as with #646, maybe this should be configurable in the not too distant future

PR #648, opened 2023-09-19
fix(local): adjust use of local dev dependenices

deer-wmde:
Currently the API image does not work in the local app env.

- Remove `MattAllan\LaravelCodeStyle` from local_providers
- Remove obsolete local check for jobs
- Remove unused IdeHelperServiceProvider
  - This one may be needed by psalm? (edit: I guess not since CI is happy with it atm)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,6 @@
     "require-dev": {
         "barryvdh/laravel-ide-helper": ">=2.8.0 <2.9.2",
         "fakerphp/faker": "^1.17",
-        "matt-allan/laravel-code-style": "^0.7.0",
         "mockery/mockery": "^1.4",
         "phpunit/phpunit": "^9.5",
         "psalm/plugin-laravel": "^1.4",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "618c3b65bfc08bcfc59eeaaf2a791965",
+    "content-hash": "7f9043a1fd97871a219a252831211715",
     "packages": [
         {
             "name": "asm89/stack-cors",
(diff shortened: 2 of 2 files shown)


app/Http/Controllers/WikiController.php
```diff
@@ -110,12 +110,6 @@ public function create(Request $request): \Illuminate\Http\Response
               'wiki_id' => $wiki->id,
             ]);
 
-            // If we are local, the dev environment wont be able to run these jobs yet, so end this closure early.
-            // TODO maybe send different jobs instead? or do this in the jobs?
-            if (App::environment() === 'local') {
```
m90, line 115: Just so I understand: why is this not needed anymore?

config/app.php
```diff
@@ -198,7 +198,6 @@
         MattAllan\LaravelCodeStyle\ServiceProvider::class,
     ],
     'local_providers' => [
-        MattAllan\LaravelCodeStyle\ServiceProvider::class,
```
m90, line 201: Can this now be removed from composer.json as well?

Or do we want to configure it to be "auto-discovered" as described here https://github.com/barryvdh/laravel-ide-helper#installation?

app/Http/Controllers/WikiController.php
deer-wmde: I'm not sure when this was needed in the first place, but nowadays the jobs that follow that check work. We just had the app env set to `production` everywhere prior to https://github.com/wmde/wbaas-deploy/commit/42a08af63af0d460d952938dce1520d76aa3c7e3#diff-bb11be128453a5ee0b102c8feaa9890809f7f925ac2dee60b620b1c351fdeb65R23, so the check wasn't even running

config/app.php
deer-wmde: I pushed a commit to remove it - I don't know if this was really used by anyone

m90 (approved):
:broom:

PR #649, opened 2023-09-19
Collect Wiki lifecycle events

m90:
Ticket https://phabricator.wikimedia.org/T345434

### DIFF
--- a/app/Jobs/UpdateWikiSiteStatsJob.php
+++ b/app/Jobs/UpdateWikiSiteStatsJob.php
@@ -6,6 +6,7 @@ use App\Wiki;
 use App\WikiSiteStats;
 use Illuminate\Support\Facades\Http;
 use Illuminate\Support\Facades\Log;
+use Illuminate\Http\Client\Pool;
 use Illuminate\Contracts\Queue\ShouldBeUnique;
 
 class UpdateWikiSiteStatsJob extends Job implements ShouldBeUnique
@@ -17,6 +18,7 @@ class UpdateWikiSiteStatsJob extends Job implements ShouldBeUnique
         foreach ($allWikis as $wiki) {
             try {
                 $this->updateSiteStats($wiki);
+                $this->updateLifecycleEvents($wiki);
             } catch (\Exception $ex) {
                 $this->job->markAsFailed();
                 Log::error(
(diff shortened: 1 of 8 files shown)


app/Jobs/UpdateWikiSiteStatsJob.php
```diff
@@ -26,6 +29,31 @@ public function handle (): void
         }
     }
 
+    private function updateLifecycleEvents (Wiki $wiki): void {
+        $responses = Http::pool(fn (Pool $pool) => [
+            $pool->as('revisions')->withHeaders(['host' => $wiki->getAttribute('domain')])->get(
```
m90, line 34: Theoretically, the value returned by this call will never change, so if we wanted to optimize for performance, we could also check if such a value has already been stored for the wiki and the only make the call if we need to.

I chose not to do this right now as we run this once a day only and performance is not an issue.

database/migrations/2023_09_19_094751_create_wiki_lifecycle_events_table.php
```diff
@@ -0,0 +1,36 @@
...
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('wiki_lifecycle_events', function (Blueprint $table) {
+            $table->id();
+            $table->timestamps();
+            $table->timestampTz('first_edited')->nullable();
```
m90, line 19: If I understand it correctly, all Wikis on wikibase.cloud run on UTC, so we could also drop the Timezone information here as it's always `+00:00`. Then again it's probably safer to already add it.

app/WikiLifecycleEvents.php
```diff
@@ -0,0 +1,28 @@
...
+
+namespace App;
+
+use Illuminate\Database\Eloquent\Factories\HasFactory;
+use Illuminate\Database\Eloquent\Model;
+
+// This class is supposed to contain information about certain events
+// in a wiki's lifecycle, e.g. the time of the last edit. Sources for these
+// points in time can be chosen as needed.
+class WikiLifecycleEvents extends Model
```
m90, line 11: Naming suggestions are welcome. I think the current one is ok, but maybe there's something better out there still.

deer-wmde: I think that naming already fits quite good :)

deer-wmde (approved):
looks very good to me overall!

PR #652, opened 2023-09-21
fix(domains): store domains in IDN format

deer-wmde:
https://phabricator.wikimedia.org/T345139#9183814

~~This PR adds the package fomvasss/laravel-punycode: https://github.com/fomvasss/laravel-punycode~~

- as of now the changes would mean that domains get stored in the [IDN format](https://en.wikipedia.org/wiki/Internationalized_domain_name) via a wrapper for the [native PHP IDN functions](https://www.php.net/manual/en/ref.intl.idn.php).
- a `domain_decoded` field gets added via an Eloquent accessor: 
  - con: depending on the implementation, it can be necessary to add it manually (like in `WikiPublicResource` (discovery feature))
  - pro: this way (storing & handling domains encoded) we avoid other instances of encoding issues  

How to test locally:
- open minikube tunnel
- get your ingress nginx IP
  - `echo $(minikube --profile minikube-wbaas kubectl -- -n kube-system get service ingress-nginx-controller -o template='{{.spec.clusterIP}}')`
- add entry in `/etc/hosts` for e.g. `10.106.231.200 xn--ento-ioa.carolinadoran.com`
- create wiki with custom domain `então.carolinadoran.com`

### DIFF
--- a/app/Http/Controllers/WikiController.php
+++ b/app/Http/Controllers/WikiController.php
@@ -19,6 +19,7 @@ use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Str;
 use Illuminate\Support\Facades\Config;
 use App\Helper\DomainValidator;
+use Fomvasss\Punycode\Facades\Punycode;
 
 class WikiController extends Controller
 {
@@ -32,8 +33,10 @@ class WikiController extends Controller
     public function create(Request $request): \Illuminate\Http\Response
     {
         $user = $request->user();
-        $submittedDomain = strtolower($request->input('domain'));
         
+        $submittedDomain = strtolower($request->input('domain'));
+        $submittedDomain = PunyCode::encode($submittedDomain);
+
         $validator = $this->domainValidator->validate( $submittedDomain );
         $isSubdomain = $this->isSubDomain($submittedDomain);
 
@@ -49,7 +52,7 @@ class WikiController extends Controller
         $dbAssignment = null;
 
         // TODO create with some sort of owner etc?
-        DB::transaction(function () use ($user, $request, &$wiki, &$dbAssignment, $isSubdomain) {
+        DB::transaction(function () use ($user, $request, &$wiki, &$dbAssignment, $isSubdomain, $submittedDomain) {
             $dbVersion = Config::get('wbstack.wiki_db_use_version');
             $wikiDbCondition = ['wiki_id'=>null, 'version'=>$dbVersion];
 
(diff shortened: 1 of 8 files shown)


app/WikiDomain.php
```diff
@@ -39,4 +40,14 @@ public function wiki(): \Illuminate\Database\Eloquent\Relations\HasOne
     {
         return $this->hasOne(Wiki::class);
     }
+
+    static public function encode($domain)
```
tarrow, line 44: No objection to sticking this in here but as you said it's interesting that we don't really seem to use this model much. Maybe we ought to think about retiring it

app/Http/Controllers/WikiController.php
```diff
@@ -185,8 +187,16 @@ public function getWikiDetailsForIdForOwner(Request $request): \Illuminate\Http\
...
       ->with('publicSettings')->first();
 
-        $res['success'] = true;
-        $res['data'] = $wiki;
+        $res = [
+            'success' => true,
+            'data'    => $wiki,
+        ];
+
+        data_set($res, 'data.domain',
```
tarrow, line 195: I'm not sure I've seen this pattern before: what's the advantage of this over just returning it unchanged?

deer-wmde: you mean the `data_set` function? It's a Laravel helper https://laravel.com/docs/8.x/helpers#method-data-set

I was under the impression that PHP is unhappy if you use unset indices in arrays but I just tested it and it seems that's only true for accessing, it seems to just set the indices if they are not there, which makes sense. I'll change it for simplicity

app/WikiDomain.php
deer-wmde: Writing a test rn - also I think maybe its own helper class would make sense

deer-wmde:
Pivoted a bit from using that Laravel package because I found out about native PHP functions and the IDN format (which is what we want, and while it's using punycode, it and it's encoding implementation is not the same, as it is specific for domains).

deer-wmde:
I implemented the unicode conversion in the UI here: https://github.com/wbstack/ui/pull/742

But I have to say that it doesn't feel right for us needing to alter the API data everywhere where domains appear. On the other hand, this would be needed then on the API side. I wonder if we can find some middle ground, maybe a new field in the API response, or a generic conversion if the value is casted to String ([Stringable interface](https://www.php.net/manual/en/class.stringable.php)).

tarrow:
Thanks for all the thoughts and discussion on this topic; I think we've now come to some pretty happy conclusion. I'm sure that all the thinking that went into this was worthwhile. Let's go down this path and see how it works. We'll want to merge this first and then follow-up with the UI fix

app/WikiDomain.php
tarrow: Sure, as we discussed; using the help makes sense if you feel we really add some clarity with the tests

PR #656, opened 2023-10-10
feat: Create EmptyWikibaseNotification Model and Notification

dati18:
https://phabricator.wikimedia.org/T344689


app/EmptyWikibaseNotification.php
```diff
@@ -0,0 +1,27 @@
...
+use Illuminate\Database\Eloquent\Factories\HasFactory;
+use Illuminate\Database\Eloquent\Model;
+
+class EmptyWikibaseNotification extends Model
+{
+    use HasFactory;
+
+    const FIELDS = [
+        'domain',
+        'first_edited',
```
m90, line 14: This data is already present in `WikiLifecycleEvents`. What's the reason for choosing to store it again in a different model?

database/migrations/2023_10_31_204320_create_wikibase_notification_sent_records_table.php
```diff
@@ -0,0 +1,33 @@
...
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('wikibase_notification_sent_records', function (Blueprint $table) {
+            $table->id();
+            $table->timestamps();
+
+            $table->string('notification_type') -> nullable();
```
m90, line 20: This table still needs a `wiki_id` foreign key so we can map it to a wiki.

See here for an example https://github.com/wbstack/api/blob/5a8bc9f2ac07d223cb5124d434f400d061b1246b/database/migrations/2023_09_19_094751_create_wiki_lifecycle_events_table.php#L22-L23

app/Console/Kernel.php
```diff
@@ -46,6 +47,8 @@ protected function schedule(Schedule $schedule)
         $schedule->job(new PollForMediaWikiJobsJob)->everyFifteenMinutes();
 
         $schedule->job(new UpdateWikiSiteStatsJob)->dailyAt('19:00');
+
+        $schedule->job(new SendEmptyWikibaseNotificationsJob)->monthly();
```
m90, line 51: I think we can do this way more often, no? Maybe daily? Probably good to check with product what they expect.

app/Jobs/SendEmptyWikibaseNotificationsJob.php
```diff
@@ -0,0 +1,42 @@
...
+use App\Wiki;
+use Carbon\CarbonImmutable;
+use Illuminate\Contracts\Queue\ShouldBeUnique;
+use Illuminate\Support\Facades\Log;
+
+class SendEmptyWikibaseNotificationsJob extends Job implements ShouldBeUnique
+{
+    public function handle (): void
+    {
+        $allWikis = Wiki::all();
```
m90, line 15: I think this would still need to load the `wikiManagers` and `wikibaseNotificationSentRecord` relations.

app/Wiki.php
```diff
@@ -87,6 +87,11 @@ public function wikiLifecycleEvents(): \Illuminate\Database\Eloquent\Relations\H
         return $this->hasOne(WikiLifecycleEvents::class);
     }
 
+    public function wikibaseNotificationSentRecord(): \Illuminate\Database\Eloquent\Relations\HasOne
+    {
+        return $this->hasOne(WikibaseNotificationSentRecord::class);
```
m90, line 92: If I understand it correctly, a Wiki can have more than one sent record, so this should be `hasMany`.

app/Jobs/SendEmptyWikibaseNotificationsJob.php
```diff
@@ -0,0 +1,42 @@
...
+                $this->sendEmptyWikibaseNotification($wiki);
+            } catch (\Exception $exception) {
+                Log::error(
+                    'Failure polling wiki '.$wiki->getAttribute('domain').' for sitestats: '.$exception->getMessage()
+                );
+            }
+        }
+    }
+
+    public function sendEmptyWikibaseNotification (Wiki $wiki): void
```
m90, line 27: Very nitpicky: the method name is a bit misleading. It sounds as if this will always send a notification, whereas in reality it might skip a lot of wikis. Maybe this can be broken down in two methods anyways: 1. check if a wiki is eligible, 2. send notification

app/Jobs/SendEmptyWikibaseNotificationsJob.php
```diff
@@ -0,0 +1,42 @@
...
+    {
+        //Calculate how many days has passed since the wikibase instance was first created
+        $createdAt = $wiki->created_at;
+        $now = CarbonImmutable::now();
+        $emptyWikibaseDays = $createdAt->diffInDays($now);
+
+        $firstEdited = $wiki->first_edited;
+
+        if ($firstEdited == null && $emptyWikibaseDays >= 30) {
```
m90, line 36: I think this also still needs to check whether we have already sent this notification. Else we'll start spamming users heavily.

app/Jobs/SendEmptyWikibaseNotificationsJob.php
```diff
@@ -0,0 +1,42 @@
...
+            $user = $wiki->wikiManagers()->first();
+            $user->notify(new EmptyWikibaseNotification($wiki->sitename));
+            $wiki->wikibaseNotificationSentRecord()->updateOrCreate(['notification_type'=>'empty_wikibase_notification']);
```
m90, line 39: I think it would make sense to collect the possible values for `notification_type` as static properties on the model class. Else this is very typo prone to just put them here and there and elswhere.

app/Jobs/SendEmptyWikibaseNotificationsJob.php
m90: It would be nice if the `30` would be configurable somehow, i.e. moved to `config`.

app/Jobs/SendEmptyWikibaseNotificationsJob.php
tarrow: from Dat and I chatting: maybe something like https://laravel.com/docs/8.x/eloquent-relationships#eager-loading would be something to look at

app/Notifications/EmptyWikibaseNotification.php
```diff
@@ -0,0 +1,61 @@
...
+{
+    /**
+     * Create a notification instance.
+     *
+     * @param string $sitename
+     * @return void
+     */
+    public function __construct(string $sitename)
+    {
+        $this->sitename = $sitename;
```
m90, line 26: This property still needs to be defined on the class itself, i.e. `private $sitename;`

app/Jobs/SendEmptyWikibaseNotificationsJob.php
m90: I didn't even know lazy loading existed.

tests/Jobs/SendEmptyWikibaseNotificationsJobTest.php
```diff
@@ -0,0 +1,45 @@
...
+use App\Notifications\EmptyWikibaseNotification;
+use App\User;
+use App\Wiki;
+use App\WikiManager;
+use Illuminate\Contracts\Queue\Job;
+use Illuminate\Foundation\Testing\RefreshDatabase;
+use Illuminate\Support\Facades\Notification;
+use Tests\TestCase;
+
+class SendEmptyWikibaseNotificationsJobTest extends TestCase
```
m90, line 15: Some additional scenarios we might want to cover in the tests:
- A wiki that has seen activity, but then was abandoned does not receive a notfication
- An empty wiki receives the notification only once
- Active wikis do not receive notifications

app/Console/Kernel.php
tarrow: Added some question to the ticket: what you are saying @m90 aligns with what I would expect but I believe wasn't the expectation of @dati18 which was that a notification would be repeatedly sent every 30 days. I added https://phabricator.wikimedia.org/T344689#9300833 to try and clarify this discrepancy in understanding

dati18: ok Evelien has answered. I will fix this now :)

app/Jobs/SendEmptyWikibaseNotificationsJob.php
dati18: COuld you give me a name suggestion?

m90: I'd break it down into two methods:

```php
$isEligible = $this->checkIfWikiIsEligible($wiki);
if ($isEligible) {
    $this->sendEmptyWikibaseNotification($wiki);
}
```

app/Jobs/SendEmptyWikibaseNotificationsJob.php
```diff
@@ -0,0 +1,54 @@
...
+    public function checkIfWikiIsOldAndEmpty(Wiki $wiki)
+    {
+        //Calculate how many days has passed since the wikibase instance was first created
+        $createdAt = $wiki->created_at;
+        $now = CarbonImmutable::now();
+        $emptyWikibaseDays = $createdAt->diffInDays($now);
+
+        $firstEdited = $wiki->first_edited;
+
+        $sentNotification = WikibaseNotificationSentRecord::where('wiki_id', $wiki->id)->get(['notification_type'])->first(); //we want not just checking the 1st one but any 'empty_wikibase_notification'
```
m90, line 38: This might be any notification. If we want to know whether a wiki has ever seen this notification yet, we can do:

```php
WikibaseNotificationSentRecord::where(['wiki_id' => $wiki->id, 'notification_type' => 'the_type'])->count() > 0
```

app/Jobs/SendEmptyWikibaseNotificationsJob.php
```diff
@@ -0,0 +1,54 @@
...
+        echo $sentNotification;
+
+        if ($firstEdited == null && $emptyWikibaseDays >= 30 && $sentNotification['notification_type'] != 'empty_wikibase_notification') {
+            return true;
```
m90, line 42: This function should also properly return `false` in case the check fails.

app/Jobs/SendEmptyWikibaseNotificationsJob.php
```diff
@@ -0,0 +1,54 @@
...
+        }
+    }
+
+    public function sendEmptyWikibaseNotification (Wiki $wiki): void
+    {
+        if ($this->checkIfWikiIsOldAndEmpty($wiki)) {
```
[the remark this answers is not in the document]
dati18:
```
foreach ($wikis as $wiki) {
            try {
                if ($this->checkIfWikiIsOldAndEmpty($wiki)) {
                    $this->sendEmptyWikibaseNotification($wiki);
                }
            } catch (\Exception $exception) {
                Log::error(
                    'Failure polling wiki '.$wiki->getAttribute('domain').' for sitestats: '.$exception->getMessage()
                );
            }
        }
```
you mean this?

database/migrations/2023_10_31_204320_create_wikibase_notification_sent_records_table.php
```diff
@@ -0,0 +1,38 @@
...
+    public function up()
+    {
+        Schema::create('wikibase_notification_sent_records', function (Blueprint $table) {
+            $table->id();
+            $table->timestamps();
+
+            $table->integer('wiki_id')->nullable()->unsigned();
+//
+//            $table->unsignedInteger('wiki_id');
+//            $table->foreign('wiki_id')->references('id')->on('wikis');
```
m90, line 23: I think we still want this foreign key?

tests/Jobs/SendEmptyWikibaseNotificationsJobTest.php
```diff
@@ -0,0 +1,80 @@
...
+        $job = new SendEmptyWikibaseNotificationsJob();
+        $job->setJob($mockJob);
+        $job->handle();
+    }
+
+    public function testEmptyWikibaseNotifications_SendNotification()
+    {
+        Notification::fake();
+        $user = User::factory()->create(['verified' => true]);
+        $wiki = Wiki::factory()->create(['created_at' => '2022-12-31 16:00:00']);
```
m90, line 37: This is a personal preference, so feel free to disregard: I think it's more easy to read if you construct a date like this:
```php
Wiki::factory()->create(['created_at' => Carbon::now()->subDays(100));
```
This way I can instantly understand the wiki is 100 days old. Right now, I need to do some mental gymnastics to figure out whether the given date is long ago or not.

database/migrations/2023_10_31_204320_create_wikibase_notification_sent_records_table.php
dati18: I commented them out while debugging, of course we need that

app/Jobs/SendEmptyWikibaseNotificationsJob.php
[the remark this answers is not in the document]
m90: Yes. exactly. Like this we get two methods that have clear responsibilities and which could even be unit tested to confirm our logic assumptions.

app/Jobs/SendEmptyWikibaseNotificationsJob.php
```diff
@@ -0,0 +1,54 @@
...
+use App\WikibaseNotificationSentRecord;
+use Carbon\CarbonImmutable;
+use Illuminate\Contracts\Queue\ShouldBeUnique;
+use Illuminate\Support\Facades\Log;
+
+class SendEmptyWikibaseNotificationsJob extends Job implements ShouldBeUnique
+{
+    public function handle (): void
+    {
+        $wikis = Wiki::with('notification')->get();
```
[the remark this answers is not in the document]
m90: Ah, I was expecting the method on the model to be called `wikibaseNotificationSendRecords` which maps clearly to the model class we load. Should we rename this so it's more obvious?

app/Jobs/SendEmptyWikibaseNotificationsJob.php
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        if ($firstEdited == null && $emptyWikibaseDays >= 30 && $sentNotificationCheck == false) {
+            return true;
+        }
+    }
+
+    public function sendEmptyWikibaseNotification (Wiki $wiki): void
+    {
+        if ($this->checkIfWikiIsOldAndEmpty($wiki)) {
```
m90, line 53: I think it would be nice to hoist this check one level up (i.e. into `handle`) so that `sendEmptyWikibaseNotification` and `checkIfWikiIsOldAndEmpty` are totally independent from each other (and also keep the promises they make in their names). Alternatively this method would need to be called `sendEmptyWikibaseNotificationIfEligible`

app/Jobs/SendEmptyWikibaseNotificationsJob.php
```diff
@@ -0,0 +1,59 @@
...
+            $user = $wiki->wikiManagers()->first();
+            $user->notify(new EmptyWikibaseNotification($wiki->sitename));
+            $wiki->wikibaseNotificationSentRecord()->updateOrCreate(['notification_type' => 'empty_wikibase_notification']);
```
m90, line 56: I'd stick all possible values for  `notification_type` on static properties of the `WikibaseNotificationSentRecord` class and the compare against these (guarding us against typos):

```php
$wiki->wikibaseNotificationSentRecord()->create(['notification_type' => WikibaseNotificationSentRecord::NOTIFICATION_TYPE_EMPTY_WIKIBASE);
```

Also, as this notification should only ever be sent once, I think it's safer to only use `create` instead of `updateOrCreate`

database/migrations/2023_10_31_204320_create_wikibase_notification_sent_records_table.php
m90, line 23: This is still needed?

tests/Jobs/SendEmptyWikibaseNotificationsJobTest.php
```diff
@@ -0,0 +1,83 @@
...
+        $job->setJob($mockJob);
+        $job->handle();
+    }
+
+    // empty wikis, that are older than 30 days, trigger a notification
+    public function testEmptyWikibaseNotifications_SendNotification()
+    {
+        Notification::fake();
+        $user = User::factory()->create(['verified' => true]);
+        $wiki = Wiki::factory()->create(['created_at' => '2022-12-31 16:00:00']);
```
m90, line 39: This is a personal preference, so feel free to disregard: I think it's easier to read if you construct a date like this:
```php
Wiki::factory()->create(['created_at' => Carbon::now()->subDays(100));
```
This way I can instantly understand the wiki is 100 days old when the test runs. Right now, I need to do some mental gymnastics to figure out whether the given date is long ago or not.

app/Jobs/SendEmptyWikibaseNotificationsJob.php
```diff
@@ -0,0 +1,59 @@
...
+use Carbon\CarbonImmutable;
+use Illuminate\Contracts\Queue\ShouldBeUnique;
+use Illuminate\Support\Facades\Log;
+use Illuminate\Support\Arr;
+
+class SendEmptyWikibaseNotificationsJob extends Job implements ShouldBeUnique
+{
+    public function handle (): void
+    {
+        $wikis = Wiki::with(['notification', 'wikiLifecycleEvents'])->get();
```
m90, line 17: I don't think preloading `notification` is needed as the lookup happens by querying `wikibase_notification_sent_records` directly?

app/Jobs/SendEmptyWikibaseNotificationsJob.php
```diff
@@ -0,0 +1,59 @@
...
+
+        if ($firstEdited == null && $emptyWikibaseDays >= 30 && $sentNotificationCheck == false) {
+            return true;
+        }
+    }
+
+    public function sendEmptyWikibaseNotification (Wiki $wiki): void
+    {
+        if ($this->checkIfWikiIsOldAndEmpty($wiki)) {
+            $user = $wiki->wikiManagers()->first();
```
m90, line 54: Product question: is the first manager automatically the most relevant? Should we maybe notify all of them?

app/Wiki.php
```diff
@@ -108,6 +113,11 @@ public function settings(): \Illuminate\Database\Eloquent\Relations\HasMany
         return $this->hasMany(WikiSetting::class);
     }
 
+    public function notification(): \Illuminate\Database\Eloquent\Relations\HasMany
+    {
+        return $this->hasMany(WikibaseNotificationSentRecord::class);
+    }
+
```
m90, line 120: This can be deleted as we already define the same relationship in `wikibaseNotificationSentRecord`

app/Wiki.php
```diff
@@ -87,6 +87,11 @@ public function wikiLifecycleEvents(): \Illuminate\Database\Eloquent\Relations\H
         return $this->hasOne(WikiLifecycleEvents::class);
     }
 
+    public function wikibaseNotificationSentRecord(): \Illuminate\Database\Eloquent\Relations\HasMany
```
m90, line 90: Not sure if there is an existing convention around it, but should this method be named `wikibaseNotificationSentRecords` as it's `HasMany`? The docs do this https://laravel.com/docs/10.x/eloquent-relationships#one-to-many

database/migrations/2023_10_31_204320_create_wikibase_notification_sent_records_table.php
```diff
@@ -0,0 +1,38 @@
...
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('wikibase_notification_sent_records', function (Blueprint $table) {
+            $table->id();
+            $table->timestamps();
+
+            $table->integer('wiki_id')->nullable()->unsigned();
```
m90, line 20: Not sure if this should be nullable. Do we do this for other foreign keys?

app/Jobs/SendEmptyWikibaseNotificationsJob.php
```diff
@@ -0,0 +1,59 @@
...
+        $firstEdited = $wiki->wikiLifecycleEvents->first_edited;
+
+        $sentNotification = WikibaseNotificationSentRecord::where('wiki_id', $wiki->id)->get(['notification_type'])->first(); //we want not just checking the 1st one but any 'empty_wikibase_notification'
+        $sentNotificationCheck = false;
+
+        if (Arr::accessible($sentNotification)) {
+            $sentNotificationCheck = ($sentNotification['notification_type'] == 'empty_wikibase_notification');
+        }
+
+        if ($firstEdited == null && $emptyWikibaseDays >= 30 && $sentNotificationCheck == false) {
```
m90, line 46: Can we maybe move the `30` to `config`?

app/Jobs/SendEmptyWikibaseNotificationsJob.php
```diff
@@ -0,0 +1,59 @@
...
+        $wikis = Wiki::with(['notification', 'wikiLifecycleEvents'])->get();
+
+        foreach ($wikis as $wiki) {
+            try {
+                $this->sendEmptyWikibaseNotification($wiki);
+            } catch (\Exception $exception) {
+                Log::error(
+                    'Failure processing wiki '.$wiki->getAttribute('domain').' for EmptyWikibaseNotification check: '.$exception->getMessage()
+                );
+            }
```
m90, line 26: As discussed in Mattermost, this should either call `fail` or let the exception bubble up.

app/Jobs/SendEmptyWikibaseNotificationsJob.php
deer-wmde: Great question, although I think its not possible for a wiki to be mapped to more than 1 WikiManager, I will add it since it's cheap and future proofs this part a bit in case this changes in the future

database/migrations/2023_10_31_204320_create_wikibase_notification_sent_records_table.php
```diff
@@ -0,0 +1,35 @@
...
+
+            $table->string('notification_type') -> nullable();
```
m90, line 22: I don't think the `notification_type` should be nullable. It's the most important piece of information in such a row, so we should enforce that it's being set.

database/migrations/2023_10_31_204320_create_wikibase_notification_sent_records_table.php
```diff
@@ -0,0 +1,35 @@
...
+class CreateWikibaseNotificationSentRecordsTable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('wikibase_notification_sent_records', function (Blueprint $table) {
```
m90, line 16: Should `wiki_id` be marked a foreign key in this schema:

```
$table->foreign('wiki_id')->references('id')->on('wikis');
```

database/migrations/2023_10_31_204320_create_wikibase_notification_sent_records_table.php
deer-wmde: it probably should not

app/Jobs/SendEmptyWikibaseNotificationsJob.php
m90: > lthough I think its not possible for a wiki to be mapped to more than 1 WikiManager,
I voiced the same assumption once, but was told it is already possible. Not sure if we ever do it though.

app/Jobs/SendEmptyWikibaseNotificationsJob.php
```diff
@@ -0,0 +1,62 @@
...
+        }
+    }
+
+    public function sendEmptyWikibaseNotification (Wiki $wiki): void
+    {
+        $wikiManagers = $wiki->wikiManagers()->get();
+
+        foreach($wikiManagers as $wikiManager) {
+            $wiki->wikibaseNotificationSentRecord()->create(['notification_type' => EmptyWikibaseNotification::class]);
+            $wikiManager->notify(new EmptyWikibaseNotification($wiki->sitename));
```
m90, line 59: This is a tough call: even when notifying multiple managers, should this only be created once? In case we create multiple, I'd maybe expect some kind of additional metadata on the record. Like this, we'd only get two mysterious entries without no further information about how they're different.

deer-wmde: as discussed elsewhere I will add a user_id field for a notification, so it's clear which user got notified

app/Jobs/SendEmptyWikiNotificationsJob.php
```diff
@@ -0,0 +1,70 @@
...
+
+        $emptyWikiNotificationCount = WikiNotificationSentRecord::where([
+            'wiki_id' => $wiki->id,
+            'notification_type' => EmptyWikiNotification::TYPE
+        ])->count();
+
+        if (
+            $firstEdited == null &&
+            $emptyWikiDays >= $emptyDaysThreshold &&
+            $emptyWikiNotificationCount == 0
```
m90, line 50: Genuine question from someone who doesn't know PHP all to well: is it safe to use `==` instead of `===` here. Any chance this could go wrong?

app/Jobs/SendEmptyWikiNotificationsJob.php
```diff
@@ -0,0 +1,70 @@
...
+            return false;
+        }
+    }
+
+    public function sendEmptyWikiNotification (Wiki $wiki): void
+    {
+        $wikiManagers = $wiki->wikiManagers()->get();
+
+        foreach($wikiManagers as $wikiManager) {
+            $wiki->wikiNotificationSentRecords()->create([
```
m90, line 63: Like this, we'd also create a record when the notfiy call below fails immediately. Should we create the record _after_ calling `notify`?

app/Notifications/EmptyWikiNotification.php
```diff
@@ -0,0 +1,67 @@
...
+use Illuminate\Support\Facades\Lang;
+use Illuminate\Support\HtmlString;
+
+/**
+ * Notification to be sent to empty Wikibase owner if the wiki stay empty longer than 30 days
+ */
+
+class EmptyWikiNotification extends Notification
+{
+    const TYPE = 'empty_wiki_notification';
```
m90, line 18: This is a very good place to put it :+1:

app/Wiki.php
```diff
@@ -108,6 +113,11 @@ public function settings(): \Illuminate\Database\Eloquent\Relations\HasMany
         return $this->hasMany(WikiSetting::class);
     }
 
+    public function notification(): \Illuminate\Database\Eloquent\Relations\HasMany
+    {
+        return $this->hasMany(WikiNotificationSentRecord::class);
+    }
+
```
m90, line 120: I think this can be deleted now?

config/wbstack.php
```diff
@@ -10,6 +10,7 @@
     'wiki_db_use_version' => env('WBSTACK_WIKI_DB_USE_VERSION', 'mw1.39-wbs1'),
     'wiki_hard_delete_threshold' => env('WBSTACK_HARD_DELETE_THRESHOLD', 30),
     'wiki_max_per_user' => env('WBSTACK_MAX_PER_USER', false),
+    'wiki_empty_notification_threshold' => env('WBSTACK_EMPTY_NOTIFICATION_THRESHOLD', 30),
```
m90, line 13: Do we still need to `intval` the env var? Else we're passing around a string.

tests/Jobs/SendEmptyWikiNotificationsJobTest.php
```diff
@@ -0,0 +1,105 @@
...
+use App\User;
+use App\Wiki;
+use App\WikiNotificationSentRecord;
+use App\WikiLifecycleEvents;
+use App\WikiManager;
+use Illuminate\Contracts\Queue\Job;
+use Illuminate\Foundation\Testing\RefreshDatabase;
+use Illuminate\Support\Facades\Notification;
+use Tests\TestCase;
+use function Amp\Promise\first;
```
m90, line 16: Is this used anywhere?

app/Jobs/SendEmptyWikiNotificationsJob.php
```diff
@@ -0,0 +1,70 @@
...
+namespace App\Jobs;
+
+use App\Notifications\EmptyWikiNotification;
+use App\Wiki;
+use App\WikiNotificationSentRecord;
+use Carbon\CarbonImmutable;
+use Illuminate\Contracts\Queue\ShouldBeUnique;
+use Illuminate\Support\Facades\Log;
+
+class SendEmptyWikiNotificationsJob extends Job implements ShouldBeUnique
```
m90, line 12: +1 on all the renaming `EmpytWikibase` -> `EmptyWiki`

m90 (commented):
Looks great now, I left one last round of comments, nothing major.

app/Jobs/SendEmptyWikiNotificationsJob.php
deer-wmde: I think in this particular case it's safe since the `count` method appears to return a casted `int` and if something database related goes wrong I'd expect an Exception

app/Jobs/SendEmptyWikiNotificationsJob.php
deer-wmde: That's a great point but I'm not sure if that is the case since the `notify` method appears to dispatch the sending, but I'm not entirely sure how it all works in detail. My thinking was the other way around: prevent a case where we send a notification but don't keep a record, resulting in potentially spamming a user. But now I'm not sure how to prevent both cases

app/Notifications/EmptyWikiNotification.php
deer-wmde: I agree, although I keep wondering if some sort of reference in the `WikiNotificationSentRecord` Model would make sense, but I wouldn't know how.

app/Wiki.php
deer-wmde: I guess so. I don't think we are using it anywhere now but I'm really getting confused with our models and relationships, how we expect them to look like, how we want to use them, etc. But those worries feel out of scope for this, I'll take that with me and remove this here

tests/Jobs/SendEmptyWikiNotificationsJobTest.php
deer-wmde: nope. removed

app/Notifications/EmptyWikiNotification.php
m90: I can't think of any way myself right now either unfortunately.

m90 (approved):
LGTM :t-rex: 

Do you still want to add the CHANGELOG entry to this PR so we can ship it when merged?

tarrow (approved):

PR #657, opened 2023-10-11
fix(batches): do not purge db connection as it can be reused by following jobs

m90:
Ticket: https://phabricator.wikimedia.org/T346245

For reasons unclear to me the failing jobs call this method in Laravel 
```php
    protected function registerBatchServices()
    {
        $this->app->singleton(BatchRepository::class, DatabaseBatchRepository::class);

        $this->app->singleton(DatabaseBatchRepository::class, function ($app) {
            return new DatabaseBatchRepository(
                $app->make(BatchFactory::class),
                $app->make('db')->connection($app->config->get('queue.batching.database')),
                $app->config->get('queue.batching.table', 'job_batches')
            );
        });
    }
```

returning a `null` connection as the value is not configured. I still don't understand why some invocations don't call this method and succeed. I also don't understand why this isn't documented at all.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # api
 
+## 8x.25.1 - 12 October 2023
+- Do not purge DB connections in Platform Summary Job
+
 ## 8x.25.0 - 21 September 2023
 - Collect timestamps for Wikis' Lifecycle
 
--- a/app/Jobs/PlatformStatsSummaryJob.php
+++ b/app/Jobs/PlatformStatsSummaryJob.php
@@ -131,9 +131,6 @@ class PlatformStatsSummaryJob extends Job
     {
         $wikis = Wiki::withTrashed()->with('wikidb')->get();
 
-        $manager->purge('mw');
-        $manager->purge('mysql');
-
         $conn = $manager->connection('mysql');
         $mwConn = $manager->connection('mw');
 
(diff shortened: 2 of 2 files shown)


m90:
It also seems this logic is gone entirely in newer versions of Laravel, so maybe it's just flawed: https://github.com/laravel/framework/blob/44360d852209d863d089952cbe435734319473d4/src/Illuminate/Bus/DatabaseBatchRepository.php

m90:
Not sure to be honest, it works on my machine, but I will dig into it.

m90:
@deer-wmde You are totally right, the config changes just makes it fail differently. Luckily it also pointed me to the proper fix (I think :tm:) for the issue, which is not purging the db connections in the PlatformStatsSummaryJob. Could you check this again?

tarrow:
I've checked out the patch and can add my anecdata that currently my queue seems to be free of errors post "fix"

tarrow:
I note there are a number of other Jobs that purge their DB:
- Jobs/CreateEmptyWikiDb.php
- Jobs/DeleteWikiDbJob.php
- Jobs/ProvisionWikiDbJob.php

I don't fully grasp the necessity of them. As I look at it I assume they are called because at somepoint there was an intention to use the actual per wikidb user rather than the `mediawiki-db-manager` for accessing the DB. There was probably then then assumption that we needed to "forget" the credentials so the wrong ones weren't accidentally reused by another bit of logic.

I wonder if we ought to remove all of these purge lines?

m90:
Strangely this only started surfacing after changing the way we handle the PlatformStatsSummaryJob, so I surmise these currently do not have any breaking effect, but I would also think it makes sense to remove them in order to future proof the setup. Will add the change to this PR still.

m90:
@tarrow Did so in aaa822a

m90:
Removing all purges seems to break wiki deletion jobs, so it was removed again from this PR.

tarrow (approved):

PR #673, opened 2023-10-18
Adjust QueryService updater batch order

m90:
Ticket https://phabricator.wikimedia.org/T333164

### DIFF
--- a/app/Http/Controllers/Backend/QsController.php
+++ b/app/Http/Controllers/Backend/QsController.php
@@ -64,8 +64,6 @@ class QsController extends Controller
                     'wiki_id' => $wikiId,
                     'entityIds' => implode(',', array_unique($entityBatch)),
                 ]);
-                // TODO to the loading on all batches at once? :D
-                $batch->load(['wiki', 'wiki.wikiQueryserviceNamespace']);
                 $newlyCreatedBatches[] = $batch;
             }
 
@@ -75,6 +73,7 @@ class QsController extends Controller
             }
 
             $oldestBatch->update(['done' => 1]);
+            $oldestBatch->load(['wiki', 'wiki.wikiQueryserviceNamespace']);
             return response([$oldestBatch]);
         });
 
--- a/app/Http/Controllers/Backend/QsController.php
+++ b/app/Http/Controllers/Backend/QsController.php
@@ -67,7 +67,7 @@ class QsController extends Controller
                 $newlyCreatedBatches[] = $batch;
             }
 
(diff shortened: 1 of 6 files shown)


app/Http/Controllers/Backend/QsController.php
```diff
@@ -58,66 +57,41 @@ public function getBatches(Request $request): \Illuminate\Http\Response
...
-            $sorted = collect($batches)->sortBy('id');
-
-            if ($sorted->isEmpty()) {
-                /**
-                 * If sorted collection is empty, then look back at the still $notDoneBatches
-                 * and just shove them in...
-                 * This should be done better.
-                 */
-                $sorted = $notDoneBatches;
+            $oldestBatch = collect($newlyCreatedBatches)->merge($notDoneBatches)->sortBy('id')->first();
```
tarrow, line 70: Nice shiny feature of collections; I like this. Super easy to read

app/Http/Controllers/Backend/QsController.php
```diff
@@ -58,66 +57,41 @@ public function getBatches(Request $request): \Illuminate\Http\Response
...
     public function markBatchesDone(Request $request): \Illuminate\Http\Response
     {
         $rawBatches = $request->input('batches');
         $batches = explode(',', $rawBatches);
-        foreach ($batches as $batch) {
-            \App\QsBatch::where('id', $batch)
-                ->update(['done' => 1]);
-        }
-
+        QsBatch::whereIn('id', $batches)->update(['done' => 1]);
```
tarrow, line 86: This looks efficient and idiomatic :)

database/factories/EventPageUpdateFactory.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+namespace Database\Factories;
+
+use App\EventPageUpdate;
+use Illuminate\Database\Eloquent\Factories\Factory;
+
+class EventPageUpdateFactory extends Factory
```
tarrow, line 8: Certainly no objection to having these factories but I was wondering if they are necessary? Does defining their default state as `[]` here do something more than just calling `Blah::create()` (like we do in the prod code

tarrow (approved):
Looks good! Happy to see this merged. Just some positive comments and one stylistic question about Factories for my understanding :)

database/factories/EventPageUpdateFactory.php
m90: That's a very good question. Up until now I blindly copied what was done before without thinking too much about it, but I guess it would work like you describe as well (i.e. factory-less).

PR #674, opened 2023-10-19
Release 8x.25.2

tarrow:
Small release to test if the new version of `create-pull-request` is functional in additional to batching the other GitHub actions changes we expect to be no-ops

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # api
 
+## 8x.25.2 - 19 October 2023
+- Bump Various Github Actions packages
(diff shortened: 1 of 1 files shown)


m90 (approved):

m90:
Instead of merging this, you might also just push a tag, see if it builds and then delete the tag (both git and Docker image) again. Fine with both.

tarrow:
cheers, I'l merge it. ideally I'd like to also ship the new images to "prove" the noop (or rather pull apart the unlikely possibility that some later change has a confounding issue with bumping these build only dependencies.

PR #676, opened 2023-10-25
Revert "Adjust QueryService updater batch order"

tarrow:
Reverts wbstack/api#673

Fails due to being incompatible with the updater. See:
https://github.com/wmde/wbaas-deploy/pull/1217


tarrow:
fix up by #677

PR #677, opened 2023-10-25
fix(qs): response is expected to be double wrapped

m90:
Fixup for #673 

Regression was introduced here https://github.com/wbstack/api/pull/673/files#diff-75c86b3ca73e3c47740d3c186ab6b9985f9aaecff21755727c2e56847d7da998L93


tarrow (approved):

tarrow (approved):
also changelog

PR #679, opened 2023-10-25
feat: introduce pending and failed states for qs batches

m90:
Ticket https://phabricator.wikimedia.org/T333163

### DIFF
--- a/app/Console/Kernel.php
+++ b/app/Console/Kernel.php
@@ -7,7 +7,7 @@ use App\Jobs\ProvisionQueryserviceNamespaceJob;
 use App\Jobs\ProvisionWikiDbJob;
 use App\Jobs\PruneEventPageUpdatesTable;
 use App\Jobs\PruneQueryserviceBatchesTable;
-use App\Jobs\RequeuePendingBatches;
+use App\Jobs\RequeuePendingBatchesJob;
 use App\Jobs\SandboxCleanupJob;
 use App\Jobs\PollForMediaWikiJobsJob;
 use App\Jobs\UpdateWikiSiteStatsJob;
@@ -35,7 +35,7 @@ class Kernel extends ConsoleKernel
         $schedule->job(new ExpireOldUserVerificationTokensJob)->hourly();
         $schedule->job(new PruneEventPageUpdatesTable)->everyFifteenMinutes();
         $schedule->job(new PruneQueryserviceBatchesTable)->everyFifteenMinutes();
-        $schedule->job(new RequeuePendingBatches)->everyFifteenMinutes();
+        $schedule->job(new RequeuePendingBatchesJob)->everyFifteenMinutes();
 
         // Sandbox
         // TODO this should maybe only be run when sandbox as a whole is loaded?
--- a/app/Console/Kernel.php
+++ b/app/Console/Kernel.php
@@ -7,6 +7,7 @@ use App\Jobs\ProvisionQueryserviceNamespaceJob;
 use App\Jobs\ProvisionWikiDbJob;
 use App\Jobs\PruneEventPageUpdatesTable;
 use App\Jobs\PruneQueryserviceBatchesTable;
+use App\Jobs\RequeuePendingBatches;
 use App\Jobs\SandboxCleanupJob;
 use App\Jobs\PollForMediaWikiJobsJob;
 use App\Jobs\UpdateWikiSiteStatsJob;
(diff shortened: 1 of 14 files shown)


m90:
Leaving this a draft until there is a companion PR in `queryservice-updater`

tests/Routes/QsBatch/QsControllerTest.php
```diff
@@ -57,41 +63,50 @@ public function testImplicitBatchCreation (): void
...
         $this->assertEquals(QsBatch::where('id', 1)->first()->done, 1);
+        $this->assertNotNull(QsBatch::where('id', 1)->first()->pending_since);
         $this->assertEquals(QsBatch::where('id', 2)->first()->done, 1);
+        $this->assertNull(QsBatch::where('id', 2)->first()->pending_since);
         $this->assertEquals(QsBatch::where('id', 3)->first()->done, 1);
+        $this->assertNull(QsBatch::where('id', 3)->first()->pending_since);
         $this->assertEquals(QsBatch::where('id', 4)->first()->done, 0);
+        $this->assertNotNull(QsBatch::where('id', 4)->first()->pending_since);
     }
     public function testMarkFailed (): void {
```
m90, line 94: This is now terribly named: "marking failed" does not set `failed => true` but just removes the `pending_since` and resets potential `done` values. Should this maybe be called `markNotDone` now?

Asking because the change would span multiple repos.

app/Console/Kernel.php
```diff
@@ -34,6 +35,7 @@ protected function schedule(Schedule $schedule)
         $schedule->job(new ExpireOldUserVerificationTokensJob)->hourly();
         $schedule->job(new PruneEventPageUpdatesTable)->everyFifteenMinutes();
         $schedule->job(new PruneQueryserviceBatchesTable)->everyFifteenMinutes();
+        $schedule->job(new RequeuePendingBatchesJob)->everyFifteenMinutes();
```
m90, line 38: Alternatively we could also have one "cleanup" type of job that does everything around QueryService batches.

database/migrations/2023_10_25_153643_qs_batches_add_pending_column.php
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('qs_batches', function (Blueprint $table) {
+            $table->timestampTz('pending_since')->nullable()->default(null);
+            $table->unsignedInteger('processing_attempts')->default(0);
```
m90, line 18: There'd be a million ways to model (extra failures table, single field), but this seemed like the most "dumb" solution to me. If you have a better idea or an opinion on this, now is the time to speak up.

tests/Jobs/RequeuePendingBatchesJobTest.php
```diff
@@ -0,0 +1,43 @@
...
+        parent::setUp();
+        QsBatch::query()->delete();
+    }
+
+    public function tearDown(): void {
+        QsBatch::query()->delete();
+        parent::tearDown();
+    }
+
+    public function testRequeue (): void {
```
m90, line 28: This could probably also test that errors are reported, but I have yet to find how to fake a global helper (i.e. there's no facade for `report` it seems)

tests/Routes/QsBatch/QsControllerTest.php
deer-wmde: Not sure if the renaming is necessary but I certainly need to ask for clarity now: so a QS batch considered `failed` would mean `batch did not call back as "done" after N requeueing attempts`, similarly to how the laravel job queue works?

database/migrations/2023_10_25_153643_qs_batches_add_pending_column.php
deer-wmde: In my head `pending`, `failed` and `done` would be possible states of one shared state here - but I can live with it not being 1:1 implemented that way, especially because I see how it makes sense to retro-fit it this way

tests/Routes/QsBatch/QsControllerTest.php
m90: Yes, this is exactly how it works.

database/migrations/2023_10_25_153643_qs_batches_add_pending_column.php
m90: Yeah, I thought about how this would be the cleanest option as well, introducing a new `state` field and then allowing one of the three values in there. As you already surmised I was worried about having to migrate all the existing items and what happens with the ones that are processed while running the migration. I'll give it another think.

app/Jobs/RequeuePendingBatchesJob.php
```diff
@@ -0,0 +1,34 @@
+<?php
```
deer-wmde, line 1: naming nitpick: I think it would be slightly more helpful if the job name would contain a reference to the query service, something like `RequeuePendingQsBatchesJob` for example

deer-wmde (approved):

PR #680, opened 2023-10-26
Add raw information to ConversionMetrics

tarrow:
Adds raw creation, first and last edit times to ConversionMetrics.

Also uses non-absolute values for engagement and abandonment time
this accounts for edge cases where we import wikis such that they have edits
before they are created on this WBStack instance

### DIFF
--- a/tests/Routes/Wiki/ConversionMetricTest.php
+++ b/tests/Routes/Wiki/ConversionMetricTest.php
@@ -1,6 +1,7 @@
 <?php
 
 namespace Tests\Routes\Wiki;
+use Carbon\Carbon;
 use Carbon\CarbonImmutable;
 use Tests\Routes\Traits\OptionsRequestAllowed;
 use Tests\TestCase;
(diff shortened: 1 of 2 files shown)


tests/Routes/Wiki/ConversionMetricTest.php
```diff
@@ -21,6 +22,8 @@ public function setUp(): void {
         Wiki::query()->delete();
         WikiSiteStats::query()->delete();
         WikiSetting::query()->delete();
+        Carbon::setTestNow(Carbon::parse('first day of October 2023'));
```
m90, line 25: I've never done this in PHP before, but do we need to "unfreeze" the time again in `tearDown` so that following tests don't get tripped up?

tarrow: right now all tests pass, but I think that is a wise suggestion

m90 (approved):

PR #681, opened 2023-11-06
fix(db): title column for event_page_update data is too short

m90:
Ticket: https://phabricator.wikimedia.org/T350401

For reviewing, the easiest way to reproduce the error is create a wiki page with a title like `ThisStringIsALotLongerThanFourteenCharactersWillItBeInsertedStill`?

### DIFF
--- a/database/migrations/2023_11_06_084718_event_page_update_title_length.php
+++ b/database/migrations/2023_11_06_084718_event_page_update_title_length.php
@@ -1,0 +1,36 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+use App\EventPageUpdate;
+
+class EventPageUpdateTitleLength extends Migration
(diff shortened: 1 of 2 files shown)


database/migrations/2023_11_06_084718_event_page_update_title_length.php
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        EventPageUpdate::query()->get()->each(function (EventPageUpdate $eventPageUpdate) {
```
m90, line 29: This is probably a _very_ inefficient way of doing, but considering we will likely never run it, I went for something that is easy to understand.

deer-wmde (approved):
good find, wiki page titles obviously weren't considered when modeling this entity. can confirm it works, as a new entry makes its way into the `event_page_updates` table when creating a wiki page with a long title (otherwise without the migration it currently wont)

PR #685, opened 2023-11-20
Limit QsBatch sizes, run batch creation in dedicated job

m90:
Ticket https://phabricator.wikimedia.org/T348256


app/Console/Kernel.php
```diff
@@ -36,6 +37,7 @@ protected function schedule(Schedule $schedule)
         $schedule->job(new ExpireOldUserVerificationTokensJob)->hourly();
         $schedule->job(new PruneEventPageUpdatesTable)->everyFifteenMinutes();
         $schedule->job(new PruneQueryserviceBatchesTable)->everyFifteenMinutes();
+        $schedule->job(new CreateQueryserviceBatchesJob)->everyMinute();
```
tarrow, line 40: This introduces reasonable lag to the system. Not that I think we shouldn't do this but we should communicate this to Eve and Valerie to tell the users. They should *know* that the mean time to hit the queryservice is like 35 seconds (30s here and 5s in the updater polling)

app/Jobs/CreateQueryserviceBatchesJob.php
```diff
@@ -0,0 +1,86 @@
...
+    public function __construct()
+    {
+        $this->entityLimit = Config::get('wbstack.qs_batch_entity_limit');
+    }
+
+    public function handle(): void
+    {
+        DB::transaction(function () {
+            // Get ID of the latest batch that we have created (or 0)
+            $maxIdBatch = QsBatch::orderBy('id', 'desc')->first();
```
tarrow, line 27: I found this a bit confusing when I was debugging last night; of course it's actually fine but I was wondering a bit how this works in cases where there are no undone batches. If you had a smart idea how to make this better I'd be all ears (add "consumed" flag to the event? or store this number in it's own place?). If not no pressure; I can't easily thing of a better way to do this

m90: I share the sentiment, but I also haven't got a very good idea about how to change this right now. We'd need some sort of "cursor", but I have zero clue where to store this.

app/Jobs/CreateQueryserviceBatchesJob.php
```diff
@@ -0,0 +1,86 @@
...
+                // If we already have a not done batch for this same wiki, then merge that into a new batch
+                foreach ($notDoneBatches as $qsBatch) {
+                    if ($qsBatch->wiki_id == $wikiId) {
+                        $entitiesOnBatch = explode(',', $qsBatch->entityIds);
+                        if (count($entitiesOnBatch) >= $this->entityLimit) {
+                            continue;
+                        }
+                        $entityBatch = array_merge($entityBatch, $entitiesOnBatch);
+                        // Delete the old batch
+                        $qsBatch->delete();
```
m90, line 69: I wonder if this is what we want. As we delete the batch and create a new one here, it will affect the sorting done in the QsController (which happens by id).

Alternatively, we could update the batch in such a case and keep the existing ordering.

app/Jobs/CreateQueryserviceBatchesJob.php
```diff
@@ -0,0 +1,86 @@
...
+
+            $notDoneBatches = QsBatch::where([
+                ['done', '=', 0], ['pending_since', '=', null], ['failed', '=', false]
+            ])->get();
+
+            // Insert the newly created batches into the table...
+            foreach ($wikiBatchesEntities as $wikiId => $entityBatch) {
+                // If we already have a not done batch for this same wiki, then merge that into a new batch
+                foreach ($notDoneBatches as $qsBatch) {
+                    if ($qsBatch->wiki_id == $wikiId) {
```
tarrow, line 62: I'm struggling a bit to parse the logic here; might just be me being slow though.

Does this code take into account the situation where there are many Events unbatched (and that length is already longer than the batch limit?) To me this only appears to prevent us merging batches and making one that is too big. If so maybe we want to account for that?

m90: Correct, this would not do anything on a batch of size 100 that already exists in the database. It would just not append further items to it. The batch of size 100 would remain in the db until processed though.

I wasn't too sure if this is in scope here, so I left it out. If we want this, I'd rather put it in a manual one-off command that does break down existing batches that have grown too big into properly sized ones.

app/Jobs/RequeuePendingQsBatchesJob.php
```diff
@@ -17,18 +18,20 @@ public function __construct() {
...
-        tap(QsBatch::where([
-            ['processing_attempts', '>=', $this->markFailedAfter],
-            ['failed', '=', false],
-        ]))
-            ->update(['failed' => true])
-            ->get()
-            ->each(function ($batch, $index) {
-                report("QsBatch with ID ".$batch->id."was marked as failed.");
-            });
+        DB::transaction(function () {
```
m90, line 21: Wrapped this in a transaction so we get locking and don't run into funny races somehow.

app/Jobs/CreateQueryserviceBatchesJob.php
m90: Added such a command in https://github.com/wbstack/api/pull/685/commits/a58d8a03b7075a4e232900e97ffa251d80bd3181

config/wbstack.php
```diff
@@ -27,4 +27,5 @@
 
     'qs_batch_pending_timeout' => env('WBSTACK_QS_BATCH_PENDING_TIMEOUT', 'PT300S'),
     'qs_batch_mark_failed_after' => intval(env('WBSTACK_QS_BATCH_MARK_FAILED_AFTER', '3')),
+    'qs_batch_entity_limit' => intval(env('WBSTACK_QS_BATCH_ENTITY_LIMIT', '10')),
```
m90, line 30: Theoretically, this would warrant a chart update, but maybe we can defer this until we actually need another value?

app/Jobs/CreateQueryserviceBatchesJob.php
m90: About "unbatched" entries: if we trust the code to do what we want it to do, this should never happen and so I'm not sure if it makes sense to add code that would cater for a case that other code shouldn't even create (not saying that it could happen though, that's correct).

app/Console/Kernel.php
tarrow: yeah, it's totally fine IMHO

app/Jobs/CreateQueryserviceBatchesJob.php
tarrow: I think I'm mostly worried about "unbatched" entries as you're calling them here. If I understand rightly (I may not) then this can happen in a few cases:

- crazy high edit rate where someone makes more than batch limit entries in under 1 min
- contention on the job runner - this might be scheduled every minute but if it doesn't actually get run due to something else in the way then this could happen?
- something happens such that *this* exact transaction fails to commit (basically what we saw yesterday) when the DB is giving some error

Am I being paranoid here? I think I could be but this exact situation caught me out in trying to fix up the site yesterday

I wouldn't block this patch on this since this is clearly better and the "hard" upper limit is clearly orders of magnitude bigger than the batchsize we are envisioning. I just forsee us possibly getting in a state where there is a backlog after an outage of some kind and we get hit by this again.

Maybe we need to add an alert on "eventTo" vs newest event id being crazy far apart?

app/Jobs/CreateQueryserviceBatchesJob.php
m90: > Perhaps a solution is to follow up and sort in the controller by oldest edit time on the QSBatch model (and not on id)

Not sure if I understand all the details here (sadly after such a long time working on this), but considering:

- the id is an incrementing int
- we never update models, but will delete and create a new one on update

wouldn't sorting on `created_at` and `id` give us the exact same results?

m90: > if we update the batch we need to rejig the "cursor" logic you mentioned in the resolved comment above I think. 

In my understanding, it's the other way around. If we update, we get a "stable" ordering of wikis in the queue, if we delete and create we have wikis being pushed back in the queue each time they see an update.

app/Jobs/CreateQueryserviceBatchesJob.php
m90: I have to admit I don't really see this happening all too often, and in case it does, we end up with multiple smaller batches than it's ideal, but this is not going to break anything. It might affect speed, but it's not going to have an effect as dramatic as what we've seen.

Or to put it that way: I'd be more worried about introducing subtle bugs, especially around accidentally changing ordering, in such a reconciliation logic than I am worried about having a non-ideal distribution of small batches.

Or am I missing some scenario in which for example a wiki with 4 batches of 3 entities each being created could hurt us?

m90: If we wanted to do this, I'd rearchitect the entire setup sth like this:

- there is one `qs_batches` table that works as previously (i.e. no limit for batch sizes) with a fixed ordering mechanism
- calling the QsController's GET method plucks the next 10 entities off the oldest item in that `qs_batches` table and turns it into a `qs_job`
- the items in `qs_job` are then subject to the pending/retry/failed logic we already have

Then again, I'm not sure it'd be worth it.

app/Jobs/CreateQueryserviceBatchesJob.php
m90: This test case might help visualizing what's currently shuffling things around: https://github.com/wbstack/api/blob/a58d8a03b7075a4e232900e97ffa251d80bd3181/tests/Jobs/CreateQueryserviceBatchesJobTest.php#L123

app/Jobs/CreateQueryserviceBatchesJob.php
```diff
@@ -0,0 +1,99 @@
...
+    public function __construct()
+    {
+        $this->entityLimit = Config::get('wbstack.qs_batch_entity_limit');
+    }
+
+    public function handle(): void
+    {
+        DB::transaction(function () {
+            // Get the latest batch that we have created (or 0)
+            $maxIdBatch = QsBatch::orderBy('eventTo', 'desc')->first();
```
m90, line 27: This works and is easy to understand, but has one terrible drawback: it makes it very hard for us to manually rewind and play back things that we missed. I start wondering if we should really just persist the value in the database, even if it means creating a table with a single column and a single row.

app/Jobs/CreateQueryserviceBatchesJob.php
```diff
@@ -0,0 +1,99 @@
...
+            foreach ($wikiBatchesEntities as $wikiId => $entityIdsFromEvents) {
+                // If we already have a not done batch for this same wiki, then merge that into a new batch
+                foreach ($notDoneBatches as $qsBatch) {
+                    if ($qsBatch->wiki_id !== $wikiId) {
+                        continue;
+                    }
+
+                    $entitiesOnBatch = explode(',', $qsBatch->entityIds);
+                    $tentativeMerge = array_unique(array_merge($entityIdsFromEvents, $entitiesOnBatch));
+                    if (count($tentativeMerge) >= $this->entityLimit) {
```
m90, line 70:
```suggestion
                    if (count($tentativeMerge) > $this->entityLimit) {
```

app/Jobs/CreateQueryserviceBatchesJob.php
tarrow: So we think that actually we do want a single column for this after much thinking

config/wbstack.php
tarrow: we can do this whenever given there is a default :)

app/Jobs/CreateQueryserviceBatchesJob.php
```diff
@@ -0,0 +1,97 @@
...
+                    continue 2;
+                }
+
+                $chunks = array_chunk($entityIdsFromEvents, $this->entityLimit);
+                foreach ($chunks as $chunk) {
+                    // Insert the new batch
+                    QsBatch::create([
+                        'done' => 0,
+                        'eventFrom' => $lastCheckpoint,
+                        'eventTo'=> $lastEventId,
```
m90, line 86: From what I understand we could now drop these columns. Not sure if we should as they could be helpful when debugging?

app/Jobs/CreateQueryserviceBatchesJob.php
```diff
@@ -0,0 +1,97 @@
...
+                    continue 2;
+                }
+
+                $chunks = array_chunk($entityIdsFromEvents, $this->entityLimit);
+                foreach ($chunks as $chunk) {
+                    // Insert the new batch
+                    QsBatch::create([
+                        'done' => 0,
+                        'eventFrom' => $lastCheckpoint,
+                        'eventTo'=> $latestEventId,
```
m90, line 86: I'm wondering if we should either:
- drop these as they are not really used anymore now
- fix them in a way that they tell us the from-to range for the batch itself. Currently the are the same for each batch created, possibly munging together the updates from multiple wikis, making it only semi-helpful for debugging

app/Http/Controllers/Backend/QsController.php
```diff
@@ -15,64 +13,14 @@ class QsController extends Controller
...
-                    'done' => 0,
-                    'eventFrom' => $batchesUpToEventId,
-                    'eventTo'=> $lastEventId,
-                    'wiki_id' => $wikiId,
-                    'entityIds' => implode(',', array_unique($entityBatch)),
-                    'pending_since' => null,
-                ]);
-                $newlyCreatedBatches[] = $batch;
-            }
+                ->orderBy('id')
```
tarrow, line 21: let's double check this doesn't need a `desc`

app/Jobs/CreateQueryserviceBatchesJob.php
tarrow: drop it we think

app/Jobs/CreateQueryserviceBatchesJob.php
```diff
@@ -0,0 +1,97 @@
...
+            $notDoneBatches = QsBatch::where([
+                ['done', '=', 0],
+                ['pending_since', '=', null],
+                ['failed', '=', false],
+            ])->get();
+
+            // Insert the newly created batches into the table...
+            foreach ($wikiBatchesEntities as $wikiId => $entityIdsFromEvents) {
+                // If we already have a not done batch for this same wiki, then merge that into a new batch
+                foreach ($notDoneBatches as $qsBatch) {
```
tarrow, line 59: maybe we can factor out this loop with a slightly more descriptive name. e.g. `attemptToInsertIntoExistingBatch` or something wittier

app/Console/Commands/QsBatches/SplitBatches.php
```diff
@@ -0,0 +1,66 @@
...
+
+namespace App\Console\Commands\QsBatches;
+
+use App\QsBatch;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Config;
+use Illuminate\Support\Facades\Log;
+use Illuminate\Console\Command;
+
+class SplitBatches extends Command
```
tarrow, line 11: wooo! we don't need to keep and maintain this anymore

database/factories/QueryServiceCheckpointFactory.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+namespace Database\Factories;
+
+use App\QueryServiceCheckpoint;
+use Illuminate\Database\Eloquent\Factories\Factory;
+
+class QueryServiceCheckpointFactory extends Factory
```
tarrow, line 8: let's rip it out

tests/Jobs/CreateQueryserviceBatchesJobTest.php
```diff
@@ -0,0 +1,205 @@
...
+
+        $existingBatch = QsBatch::where(['wiki_id' => 99])->first();
+
+        $this->assertNotNull($existingBatch);
+        $this->assertEquals($existingBatch->entityIds, 'Q34,Q99,Q100');
+        $this->assertEquals(3, QsBatch::query()->count());
+    }
+
+    function testBigBatches(): void
+    {
```
tarrow, line 96: mock config here maybe?

tarrow: or in the setup

app/Http/Controllers/Backend/QsController.php
m90: Double checked: It's fine as is.

app/Jobs/CreateQueryserviceBatchesJob.php
```diff
@@ -0,0 +1,108 @@
...
+                $latestEventId = $event->id;
+            }
+        }
+
+        return $newEntitiesFromEvents;
+    }
+
+    private function tryToAppendEntitesToExistingBatches(array $entityIdsFromEvents, int $wikiId): bool
+    {
+        $notDoneBatches = QsBatch::where([
```
m90, line 70: I moved this query a level down which means:
- we now do one query per wiki (i.e. more SQL queries being run)
- less manual filtering of wikis that don't apply

In the end it's probably a bit slower, but considering the query is not really complicated I think the fact that it lets us remove yet another if condition in a for loop is worth the performance hit. Let me know if you think otherwise.

app/Jobs/CreateQueryserviceBatchesJob.php
```diff
@@ -0,0 +1,113 @@
...
+                $this->createNewBatches($entityIdsFromEvents, $wikiId);
+            }
+
+            QsCheckpoint::set($this->getLatestEventId($lastCheckpoint));
+        });
+    }
+
+    private function getLatestEventId(int $lastCheckpoint): int
+    {
+        $next = EventPageUpdate::where(
```
m90, line 45: This also means we trade an additional DB query against removing some slightly brittle looping logic.

app/Jobs/CreateQueryserviceBatchesJob.php
m90: This went totally out of hand and I refactored basically EVERYTHING :/

I tried to do this in atomic commits though, so if you feel like we should play it safe and drop any of these refactorings (or all of them) please lmk and we shall do so.

app/Jobs/CreateQueryserviceBatchesJob.php
```diff
@@ -0,0 +1,107 @@
...
+                    if ($success) {
+                        continue;
+                    }
+                    $this->createNewBatches($entityIdsFromEvents, $wikiId);
+                } catch (\Exception $ex) {
+                    Log::error(
+                        'Failed to create batches '.implode(',', $entityIdsFromEvents).' for wiki with id '.$wikiId.': '.$ex->getMessage()
+                    );
+                    $this->fail($ex);
+                }
```
m90, line 43: Not sure about retrying vs. not failing all wikis on one naughty wiki here.

If we wanted the ideal behavior of:
- one naughty wiki does not block following wikis
- a failing wiki will be retried so we're sure this is not sth intermittent

we'd need to instead fan out each wiki into it's own job.

Not sure.

app/Jobs/CreateQueryserviceBatchesJob.php
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    public function handle(): void
+    {
+        DB::transaction(function () {
+            $latestCheckpoint = QsCheckpoint::get();
+
+            [$newEntities, $latestEventId] = $this->getNewEntities($latestCheckpoint);
+            foreach ($newEntities as $wikiId => $entityIdsFromEvents) {
+                try {
```
m90, line 32: Not sure about retrying vs. not failing all wikis on one naughty wiki here.

If we wanted the ideal behavior of:

- one naughty wiki does not block following wikis
- a failing wiki will be retried so we're sure this is not sth intermittent

we'd need to instead fan out each wiki into it's own job.

Not sure.

m90:
Companion chart update is here https://github.com/wbstack/charts/pull/132

app/Jobs/CreateQueryserviceBatchesJob.php
tarrow: I'm fine with this as it is; in general I would actually be "ok" with no try catch here because all of this work and logic is really just internal bookeeping inside the platform API.

Any failures here are not going to be due to some transient issue talking to a remote service over the network but are actually probably due to logical oversight in this one codebase.

tarrow (approved):

PR #686, opened 2023-11-20
fix(jobs): empty wiki notifications

deer-wmde:
https://phabricator.wikimedia.org/T344689#9340437

The job fails on wikis that got no `wikiLifecycleEvents` records. Such wikis won't get queried in the first place with this code.


app/Jobs/SendEmptyWikiNotificationsJob.php
```diff
@@ -13,7 +13,9 @@ class SendEmptyWikiNotificationsJob extends Job implements ShouldBeUnique
 {
     public function handle (): void
     {
-        $wikis = Wiki::with(['wikiLifecycleEvents'])->get();
+        $wikis = Wiki::with(['wikiLifecycleEvents'])
+            ->has('wikiLifecycleEvents')
```
m90, line 17: This is very nice compared to some null check, good to know the option exists.

m90 (approved):

PR #687, opened 2023-11-21
fix(logging): add logging env vars to Makefile

deer-wmde:
This adds the right log channel and level to the Makefile for enabling error output during running tests (deliberately not changed in `phpunit.xml` to still hide this output/noise in CI)


m90 (approved):
Nice find

PR #688, opened 2023-11-27
fix(qs): deduplicate entity ids before creating batches, skip deleted wikis

m90:
Ticket https://phabricator.wikimedia.org/T348256

Three fixups for #685 

1. Deduplicate entity ids before creating new batches
2. Do not create batches for wikis that have been deleted (the updater is unable to process these)
3. Increment processing attempts when requeuing a batch so unhandled exceptions in the updater are considered a failed attempt

### DIFF
--- a/app/Jobs/RequeuePendingQsBatchesJob.php
+++ b/app/Jobs/RequeuePendingQsBatchesJob.php
@@ -31,7 +31,10 @@ class RequeuePendingQsBatchesJob extends Job
 
             $threshold = Carbon::now()->subtract(new \DateInterval($this->pendingTimeout));
             QsBatch::where([['pending_since', '<>', null], ['pending_since', '<', $threshold]])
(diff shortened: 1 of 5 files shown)


app/Jobs/RequeuePendingQsBatchesJob.php
```diff
@@ -31,7 +31,10 @@ public function handle(): void
 
             $threshold = Carbon::now()->subtract(new \DateInterval($this->pendingTimeout));
             QsBatch::where([['pending_since', '<>', null], ['pending_since', '<', $threshold]])
-                ->update(['pending_since' => null, 'done' => 0]);
```
m90, line 34: The current state would retry a batch that makes the updater fail with an unhandled exception (i.e. malformed payload) forever, so requeueing needs to increment the couter as well.

deer-wmde (approved):

PR #691, opened 2023-11-28
fix(qs): raise timeout for batches job

m90:
After deploying https://phabricator.wikimedia.org/T348256 to production I noticed the job times out here.


m90:
Self merging this so I can unblock production

PR #692, opened 2023-11-28
fix(qs): hard increase memory_limit for qs batches job

m90:
The job is stil crashing

```
[2023-11-28 07:27:33] production.ERROR: Allowed memory size of 134217728 bytes exhausted (tried to allocate 8388616 bytes) {"exception":"[object] (Symfony\\Component\\ErrorHandler\\Error\\FatalError(code: 0): Allowed memory size of 134217728 bytes exhausted (tried to allocate 8388616 bytes) at /var/www/html/vendor/laravel/framework/src/Illuminate/Database/Connection.php:375)
```
This is probably the wrong solution, but maybe it helps in unblocking production

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # api
 
+## 8x.28.5 - 28 November 2023
+- Hard increase memory_limit for QsBatches job
+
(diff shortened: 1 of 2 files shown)


m90:
Self merging to unblock production

PR #696, opened 2023-11-28
Ensure that report helper is called when a batch is failed

m90:
When deploying #685 I noticed failed batches would not be reported.

It seems `tap` does not work as I thought it would be and instead of giving us the updated models down the chain, it would re-evaluate the query, yielding 0 results on `->get()` as the update just happened and nothing matches the query anymore.

Instead, we can query for candidate ids, update these in a second query and then report.

### DIFF
--- a/app/Jobs/RequeuePendingQsBatchesJob.php
+++ b/app/Jobs/RequeuePendingQsBatchesJob.php
@@ -18,20 +18,37 @@ class RequeuePendingQsBatchesJob extends Job
 
     public function handle(): void
     {
-        $failedBatches = [];
-        DB::transaction(function () use (&$failedBatches) {
-            $failedBatches = QsBatch::where([
-                ['processing_attempts', '>=', $this->markFailedAfter],
-                ['failed', '=', false],
-            ])->get()->pluck('id');
-            QsBatch::whereIn('id', $failedBatches)->update(['failed' => true]);
-        });
+        $failedBatches = $this->markBatchesFailed();
         foreach ($failedBatches as $batchId) {
             report("QsBatch with ID ".$batchId."was marked as failed.");
         }
 
+        $this->requeueStalledBatches();
+    }
+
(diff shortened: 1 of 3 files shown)


app/Jobs/RequeuePendingQsBatchesJob.php
```diff
@@ -18,23 +18,23 @@ public function __construct() {
...
-                );
+            ])->get()->pluck('id');
+            QsBatch::whereIn('id', $failedBatches)->update(['failed' => true]);
         });
+        foreach ($failedBatches as $batchId) {
+            report("QsBatch with ID ".$batchId."was marked as failed.");
+        }
+
+        $threshold = Carbon::now()->subtract(new \DateInterval($this->pendingTimeout));
+        QsBatch::where([['pending_since', '<>', null], ['pending_since', '<', $threshold]])
```
m90, line 34: There's no need for this to happen in the transaction above as it's a sungle `UPDATE` statement.

app/Jobs/RequeuePendingQsBatchesJob.php
```diff
@@ -18,23 +18,23 @@ public function __construct() {
...
-            $threshold = Carbon::now()->subtract(new \DateInterval($this->pendingTimeout));
-            QsBatch::where([['pending_since', '<>', null], ['pending_since', '<', $threshold]])
-                ->increment(
-                    'processing_attempts', 1,
-                    ['pending_since' => null, 'done' => 0]
-                );
+            ])->get()->pluck('id');
+            QsBatch::whereIn('id', $failedBatches)->update(['failed' => true]);
         });
+        foreach ($failedBatches as $batchId) {
```
m90, line 29: I moved this out of the transaction so we know everything has been comitted before we send out the errors.

tarrow (approved):
Thanks, this was super easy to read and review.

PR #698, opened 2023-11-29
Requeueing should not select failed batches

m90:
Currently, failed jobs will still be selected by the query for requeuing, creating an incorrect number of processing attempts.

This has no user facing effect, but can be confusing when debugging.

### DIFF
--- a/app/Jobs/RequeuePendingQsBatchesJob.php
+++ b/app/Jobs/RequeuePendingQsBatchesJob.php
@@ -37,7 +37,10 @@ class RequeuePendingQsBatchesJob extends Job
                 ->get()
                 ->pluck('id')
                 ->toArray();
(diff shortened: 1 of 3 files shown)


app/Jobs/RequeuePendingQsBatchesJob.php
```diff
@@ -48,10 +51,11 @@ private function requeueStalledBatches(): void
         QsBatch::where([
             ['pending_since', '<>', null],
             ['pending_since', '<', $threshold],
+            ['failed', '=', false],
         ])
             ->increment(
                 'processing_attempts', 1,
-                ['pending_since' => null, 'done' => 0]
+                ['pending_since' => null, 'done' => false]
```
m90, line 58: This is just a stylistic change so that the class always uses `true|false` instead of `true|false|0|1`

tarrow (approved):

PR #701, opened 2023-12-04
feat(k8s): add job that can spawn one off queryservice-updater

m90:
Ticket https://phabricator.wikimedia.org/T352657

Connects https://github.com/wbstack/queryservice-updater/pull/143

Note that this PR does only provide the Job class but does not yet create such jobs. This shall happen in a dedicated PR (#704) which introduces a command for creating jobs for all entities in all wikis.

### DIFF
--- a/app/Jobs/ProcessMediaWikiJobsJob.php
+++ b/app/Jobs/ProcessMediaWikiJobsJob.php
@@ -7,6 +7,7 @@ use Illuminate\Contracts\Queue\ShouldQueue;
 use Illuminate\Contracts\Queue\ShouldBeUnique;
 use Illuminate\Queue\InteractsWithQueue;
 use Illuminate\Support\Facades\Log;
+use Illuminate\Support\Facades\Config;
 use Maclof\Kubernetes\Client;
 use Maclof\Kubernetes\Models\Job as KubernetesJob;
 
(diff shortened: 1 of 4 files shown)


app/Jobs/SpawnQueryserviceUpdaterJob.php
```diff
@@ -0,0 +1,115 @@
...
+                                ],
+                            ]
+                        ],
+                        'restartPolicy' => 'Never'
+                    ]
+                ]
+            ]
+        ]);
+
+        $job = $kubernetesClient->jobs()->apply($jobSpec);
```
m90, line 98: Should this be `create` instead (making it fail when the job already exists). Not sure.

tarrow: probably depends on if we think jobs should be unique per wiki or not (e.g. in the event one fails and we try again)

m90: Maybe we can decide this when implementing the command. Right now I would think it doesn't matter, but then I don't know which retry mechanisms we come up with lateron.

app/Jobs/ProcessMediaWikiJobsJob.php
```diff
@@ -20,7 +21,7 @@ class ProcessMediaWikiJobsJob implements ShouldQueue, ShouldBeUnique
     public function __construct (string $wikiDomain)
     {
         $this->wikiDomain = $wikiDomain;
-        $this->jobsKubernetesNamespace = env('API_JOB_NAMESPACE', 'api-jobs');
+        $this->jobsKubernetesNamespace = Config::get('wbstack.api_job_namespace');
```
m90, line 24: This is a drive by fix for a mistake made earlier on this year. It will works as before, but we probably want to update the chart at some point as well to use `WBSTACK_` env vars throughout.

deer-wmde (approved):
Works nicely - I encountered some error last week when trying it out but I guess that was some unlucky constellation of my local cluster, today it worked just fine.

~~One thing I noticed that maybe we can improve (but rather related to https://github.com/wbstack/queryservice-updater/pull/143 I think) is that the Updater seems to go through 100 loops even if not necessary~~ just had an old QS Updater image

PR #704, opened 2023-12-11
Add command for rebuilding QueryService data

m90:
Ticket https://phabricator.wikimedia.org/T352662

~~Requires #701 so the command can stop dispatching noop jobs~~


app/QueryserviceNamespace.php
```diff
@@ -27,6 +28,7 @@
  */
 class QueryserviceNamespace extends Model
 {
+    use HasFactory;
```
m90, line 31: There seems to be a reason why people use factories: in the test i need to fill `wiki_id` which isn't declared fillable. On a factory model _everything_ is fillable, which seems preferrable over making the field fillable, just so I can write a test.

config/wbstack.php
```diff
@@ -28,4 +28,7 @@
     'qs_batch_pending_timeout' => env('WBSTACK_QS_BATCH_PENDING_TIMEOUT', 'PT300S'),
     'qs_batch_mark_failed_after' => intval(env('WBSTACK_QS_BATCH_MARK_FAILED_AFTER', '3')),
     'qs_batch_entity_limit' => intval(env('WBSTACK_QS_BATCH_ENTITY_LIMIT', '10')),
+
+    'qs_rebuild_chunk_size' => intval(env('WBSTACK_QS_REBUILD_CHUNK_SIZE', '50')),
```
m90, line 32: Not sure this is the best place to put it, considering this is a command. Should it maybe be a command line param with a default value instead?

app/Console/Commands/RebuildQueryserviceData.php
```diff
@@ -0,0 +1,153 @@
...
+                'name' => WikiSetting::wwExtEnableWikibaseLexeme
+            ],
+        )->first();
+        $hasLexemesEnabled = $lexemesSetting !== null && $lexemesSetting->value === '1';
+        $lexemes = $hasLexemesEnabled
+            ? $this->fetchPagesInNamespace($wiki->domain, self::NAMESPACE_LEXEME)
+            : [];
+
+        $merged = array_merge($items, $properties, $lexemes);
+        return $this->stripPrefixes($merged);
```
deer-wmde, line 85: I'm intrigued that this syntax works when the method is declared `static`; Until now I thought in that case you need to write `self::methodName()` or `className::methodName()`

app/Console/Commands/RebuildQueryserviceData.php
```diff
@@ -0,0 +1,153 @@
...
+                    'Failed to get prerequisites for enqueuing wiki '.$wiki->domain.', will not dispatch jobs.'
+                );
+                Log::error($ex);
+                $exitCode = 1;
+                break;
+            }
+
+            $entityChunks = array_chunk($entities, $this->chunkSize);
+            foreach ($entityChunks as $entityChunk) {
+                dispatch(
```
deer-wmde, line 55: This is probably hard to circumvent because of the nature of this batching technique but I think it can get weird if you have a large wiki and the batch job fails for some reason for some (or all) of the batches - suddenly you have to deal with a lot of batches.

For example right now locally for me somethingTM is wrong, 4 jobs get dispatched, all of them fail and it gets hard to look through the logs. Still I think this might be the desired behaviour, I just wanted to flag this

app/Console/Commands/RebuildQueryserviceData.php
```diff
@@ -0,0 +1,152 @@
...
+
+        foreach ($wikis as $wiki) {
+            try {
+                $entities = $this->getEntitiesForWiki($wiki);
+                $sparqlUrl = $this->getSparqlUrl($wiki);
+            } catch (\Exception $ex) {
+                Log::error(
+                    'Failed to get prerequisites for enqueuing wiki '.$wiki->domain.', will not dispatch jobs.'
+                );
+                Log::error($ex);
```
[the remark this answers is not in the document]
m90: Ah, I tried `Log::error("Foo", $ex)` which did not work, but this is nice to know, will change.

app/Console/Commands/RebuildQueryserviceData.php
m90: Are there any downsides to using `$this` here? Happy to change it to `self`, even if it's just for being more idiomatic.

app/Console/Commands/RebuildQueryserviceData.php
m90: I was wondering the same, but couldn't come up with something really good.

Two things I thought of:
- introduce a failure allowance and instead of continuing, bail when too many wikis have failed (this would happen when fetching the data above though)
- Introduce a `limit` and `offset` parameter and allow running the command in batches instead of for ALL wikis at once

app/Console/Commands/RebuildQueryserviceData.php
deer-wmde: I couldn't find any real downsides, gotta love this SO answer, I wasn't aware the other way around is also possible sometimes apparently...
> This means that -> is mostly used to access instance members (though it can also be used to access static members, such usage is discouraged), while :: is usually used to access static members (though in a few special cases, it's used to access instance members).

https://stackoverflow.com/questions/3173501/whats-the-difference-between-double-colon-and-arrow-in-php

app/Console/Commands/RebuildQueryserviceData.php
deer-wmde: yeah seems tricky... fwiw I'm totally okay with how it is right now, as long as we're aware how it works we should be fine

deer-wmde (approved):

PR #706, opened 2023-12-12
Add QuestyCaptcha WikiSettings to API

tarrow:
Adds two new types of WikiSetting to the platform api to manage using QuestyCaptcha.

It adds a boolean setting for turning the feature on and off as well as a totally independent setting for the questions.
This enables wiki administrators to store and curate their questions even when the setting is disabled.

It does not provide any default questions; this has been done in the UI.

This follows the existing pattern for equivalent entities and therefore stores the settings map as JSON. This can be mildly confusing since there in then double encoded JSON in the payloads to and from this API.

Validation for the content of the questions is also minimal; there is no escaping of html content in the questions since, as I understand it, this is handled in MW[1].

This can be merged independently of wbstack/mediawiki#405 but if you care to test them then it may be easier to do in parallel.

[1] https://doc.wikimedia.org/mediawiki-core/1.27.1/php/classHtml.html#a1c74fee14762ec4d50e09a94c94327ef

### DIFF
--- a/app/Http/Controllers/WikiSettingController.php
+++ b/app/Http/Controllers/WikiSettingController.php
@@ -2,6 +2,7 @@
 
 namespace App\Http\Controllers;
 
+use App\Rules\SettingCaptchaQuestions;
 use App\Rules\SettingWikibaseManifestEquivEntities;
 use App\WikiManager;
 use App\WikiSetting;
@@ -26,6 +27,8 @@ class WikiSettingController extends Controller
             'wwWikibaseStringLengthMultilang' => ['required', 'integer', 'between:250,2500'],
             'wikibaseFedPropsEnable' => ['required', 'boolean'],
             'wikibaseManifestEquivEntities' => ['required', 'json', new SettingWikibaseManifestEquivEntities()],
+            'wwUseQuestyCaptcha' => ['required', 'boolean'],
+            'wwCaptchaQuestions' => [ 'required', 'json', new SettingCaptchaQuestions()]
         ];
     }
 
--- a/app/Rules/SettingCaptchaQuestions.php
+++ b/app/Rules/SettingCaptchaQuestions.php
@@ -1,0 +1,59 @@
+<?php
+
+namespace App\Rules;
+
+use Illuminate\Contracts\Validation\Rule;
+
+class SettingCaptchaQuestions implements Rule
+{
+    /**
+     * Determine if the validation rule passes.
+     *
+     * @param  string  $attribute
+     * @param  mixed  $value
+     * @return bool
+     */
+    public function passes($attribute, $value)
+    {
+        $value = json_decode($value, true);
+
+        if ($value === null) {
+            return false;
+        }
+
+        foreach (array_keys($value) as $question) {
+            if (!is_string($question)) {
+                return false;
+            }
+            if (strlen($question) > 200) {
+                return false;
+            }
+        }
+
+        foreach (array_values($value) as $answers) {
+            if (!is_array($answers)) {
+                return false;
+            }
+            if (count($answers) === 0) {
+                return false;
+            }
+            foreach ($answers as $answer) {
+                if (!is_string($answer)) {
+                    return false;
+                }
+                if (strlen($answer) > 200) {
+                    return false;
+                }
+            }
+        }
+        return true;
+    }
+
+    /**
+     * Get the validation error message.
+     */
+    public function message(): string
+    {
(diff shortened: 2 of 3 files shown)


app/Rules/SettingCaptchaQuestions.php
```diff
@@ -0,0 +1,60 @@
...
+                return false;
+            }
+            if (count($answers) === 0) {
+                return false;
+            }
+            foreach ($answers as $answer) {
+                if (!is_string($answer)) {
+                    return false;
+                }
+                if (strlen($question) > 200) {
```
m90, line 45:
```suggestion
                if (strlen($answer) > 200) {
```

app/Rules/SettingCaptchaQuestions.php
```diff
@@ -0,0 +1,60 @@
+<?php
```
m90, line 1: Should we also have a test case for this Rule. I think it'd be pretty straight forward to set up and will help future generations in understanding what rules a set of questions is expected to adhere to.

app/Http/Controllers/WikiSettingController.php
```diff
@@ -26,6 +27,8 @@ private function getSettingValidations(): array
             'wwWikibaseStringLengthMultilang' => ['required', 'integer', 'between:250,2500'],
             'wikibaseFedPropsEnable' => ['required', 'boolean'],
             'wikibaseManifestEquivEntities' => ['required', 'json', new SettingWikibaseManifestEquivEntities()],
+            'wwUseQuestyCaptcha' => ['required', 'boolean'],
+            'wwCaptchaQuestions' => [ 'required', 'json', new SettingCaptchaQuestions()]
```
m90, line 31:
```suggestion
            'wwCaptchaQuestions' => ['required', 'json', new SettingCaptchaQuestions()]
```

app/Rules/SettingCaptchaQuestions.php
```diff
@@ -0,0 +1,60 @@
...
+        }
+        return true;
+    }
+
+    /**
+     * Get the validation error message.
+     */
+    public function message(): string
+    {
+        return 'Value must be JSON mapping of questions to an array of answers';
```
m90, line 58: Should this also mention the 200 characters limit we set for questions and answers?

m90 (commented):
Left some comments, there's one typo and I think :tm: we should have tests for the validation.

tests/Routes/Wiki/SettingUpdateTest.php
```diff
@@ -71,6 +71,15 @@ public function provideValidSettings()
...
         yield ['wikibaseManifestEquivEntities', $validProps, $validProps];
+
+        yield ['wwUseQuestyCaptcha', '1', '1'];
+        yield ['wwUseQuestyCaptcha', '0', '0'];
+        $validCaptchaQuestions = json_encode([
+            "How many vowels are in this question?" => ['12', 'twelve'],
+            "What is the chemical formula of water" => ['H2O'],
+            "2 + 4 = ?" => ['6', 'six']
+        ]);
+        yield ['wwCaptchaQuestions', $validCaptchaQuestions, $validCaptchaQuestions ];
```
m90, line 82:
```suggestion
        yield ['wwCaptchaQuestions', $validCaptchaQuestions, $validCaptchaQuestions];
```

app/Rules/SettingCaptchaQuestions.php
tarrow: Thanks! I'm a fool and clearly having validation tests for this would have caught my mistake.

tests/Routes/Wiki/SettingUpdateTest.php
tarrow: should probably add some linting for this thing ;)

app/Rules/SettingCaptchaQuestions.php
tarrow: Perhaps we should; perhaps 200 is even too short. I'll add something for now though

app/Rules/SettingCaptchaQuestions.php
tarrow: I'm actually even more of a fool; apparently I'd failed to push the right amended commit

app/Rules/SettingCaptchaQuestions.php
tarrow: I was torn on whether or not the test coverage for the controller is sufficient or if it makes sense to also add a test for the rule. I could definitely be convinced but as I was writing tests for the rule it felt mostly like I was copy/pasting the controller tests so I've omitted it for now

m90 (approved):

PR #708, opened 2023-12-13
fix: edit items in place when stripping prefixes, do not use regex

m90:
Running the command in staging made the command exit 

```
In RebuildQueryserviceData.php line 159:
                                                                                       
  Allowed memory size of 134217728 bytes exhausted (tried to allocate 33554440 bytes)  
                                                                                       
```

I'm not entirely sure what exactly consumes so much memory, but this PR tries to mitigate this by doing two things:
- edit the array of IDs in place
- use `split` instead of `preg_replace`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,13 @@
 # api
 
-## 8x.28.11 - 12 December 2023
-- Batches that are done shall never be marked failed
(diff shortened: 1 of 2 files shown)


deer-wmde (approved):
exciting, let's see what happens! :crossed_fingers:

PR #709, opened 2023-12-14
feat: Allow passing a queue name to the worker

m90:
Ticket https://phabricator.wikimedia.org/T352662

This change is backwards compatible and could be deployed without any additional chart changes.

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -18,7 +18,7 @@ if [ "$role" = "app" ]; then
 elif [ "$role" = "queue" ]; then
 
     echo "Running the $queue_name queue..."
-    php /var/www/html/artisan queue:work --verbose --tries=5 --timeout=90 --queue=$queue_name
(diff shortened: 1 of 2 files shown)


start.sh
```diff
@@ -16,8 +17,8 @@ if [ "$role" = "app" ]; then
 
 elif [ "$role" = "queue" ]; then
 
-    echo "Running the queue..."
-    php /var/www/html/artisan queue:work --verbose --tries=5 --timeout=90
+    echo "Running the $queue_name queue..."
+    php /var/www/html/artisan queue:work --verbose --tries=5 --timeout=90 --queue=$queue_name
```
deer-wmde, line 21:
```suggestion
    php /var/www/html/artisan queue:work --verbose --tries=5 --timeout=90 --queue="$queue_name"
```

most likely won't matter but I've been bitten by escaping so many times, I can't not recommend it :D

deer-wmde (approved):

PR #710, opened 2023-12-14
feat: Dispatch jobs for rebuilding Queryservice data to non-default queue

m90:
Ticket: https://phabricator.wikimedia.org/T352662

Connects https://github.com/wmde/wbaas-deploy/pull/1322

### DIFF
--- a/app/Console/Commands/RebuildQueryserviceData.php
+++ b/app/Console/Commands/RebuildQueryserviceData.php
@@ -5,6 +5,7 @@ namespace App\Console\Commands;
 use Illuminate\Console\Command;
 use Illuminate\Support\Facades\Http;
 use Illuminate\Support\Facades\Log;
+use Illuminate\Support\Facades\Queue;
 use App\Wiki;
 use App\WikiSetting;
 use App\QueryserviceNamespace;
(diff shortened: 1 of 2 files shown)


app/Console/Commands/RebuildQueryserviceData.php
```diff
@@ -16,18 +17,20 @@ class RebuildQueryserviceData extends Command
     private const NAMESPACE_PROPERTY = 122;
     private const NAMESPACE_LEXEME = 146;
 
-    protected $signature = 'wbs-qs:rebuild {--domain=*} {--chunkSize=50} {--sparqlUrlFormat=http://queryservice.default.svc.cluster.local:9999/bigdata/namespace/%s/sparql}';
+    protected $signature = 'wbs-qs:rebuild {--domain=*} {--chunkSize=50} {--queueName=manual-intervention} {--sparqlUrlFormat=http://queryservice.default.svc.cluster.local:9999/bigdata/namespace/%s/sparql}';
```
deer-wmde, line 20: does this default value mean we have to make sure to have a api-queue with that name running in case no queue is specified?

m90: Yes, this is exactly how it works. Would you think `default` is a better default here?

deer-wmde: ok thanks for clarifying. hm I'm not sure, depends on how we will want to use the default queue in the future. If we will still use it for day-to-day stuff that gets automatically queued then I'm in favor of this approach because it makes it so visible if jobs were kicked off manually :+1:

deer-wmde (approved):

PR #718, opened 2024-01-10
Update Laravel to version 9

m90:
Ticket https://phabricator.wikimedia.org/T341796


app/Jobs/DeleteWikiDbJob.php
```diff
@@ -107,6 +107,9 @@ public function handle( DatabaseManager $manager )
                 $replacedCount = 0;
                 $tableWithoutPrefix = str_replace($wikiDB->prefix . '_', '', $table, $replacedCount );
                 if ($replacedCount !== 1) {
+                    /**
+                     * @psalm-suppress InvalidCast
```
m90, line 111: I really tried hard to understand these warnings but I could not make any sense of them so I suppressed them for now. If anyone has a hunch why they might show up, please lmk.

app/Http/Controllers/Backend/QsController.php
```diff
@@ -34,21 +34,21 @@ public function getBatches(Request $request): \Illuminate\Http\Response
 
     public function markBatchesDone(Request $request): \Illuminate\Http\Response
     {
-        $batches = (array) $request->json()->get('batches');
+        $batches = (array) $request->json()->all('batches');
```
m90, line 37: This was wrong usage before and worked mostly by accident. Now, Symfony will throw an error on the wrong usage, so I was able to fix it.

m90:
> Also I wonder if we should keep certain files in sync that change https://laravel.com/docs/9.x/upgrade#miscellaneous (and https://github.com/laravel/laravel/compare/8.x...9.x). The guide says that many changes are not required though.

I didn't understand this repo was an application skeleton. I thought it contained the framework itself and was puzzled as why I would need to read all commit messages ever. I'll have a look at the changeset if I can find anything that might be helpful for us.

deer-wmde:
> I think this is benefitial actually. We rarely use the Laravel Http client anyways, and in the cases we do, If we have anything the API calls and a single call takes longer than 30s, we might wanna know about it by it failing.

That makes sense, thanks.

Also thanks for looking into the skeleton changes - I haven't done an laravel upgrade yet but now it looks pretty good to me!

deer-wmde (approved):

PR #719, opened 2024-01-10
feat: use s3 driver for storing files in gcp bucket

m90:
Ticket https://phabricator.wikimedia.org/T354744

### DIFF
--- a/config/filesystems.php
+++ b/config/filesystems.php
@@ -55,33 +55,18 @@ $fileSystems = [
             'visibility' => 'public',
         ],
 
-//        's3' => [
-//            'driver' => 's3',
-//            'key' => env('AWS_ACCESS_KEY_ID'),
-//            'secret' => env('AWS_SECRET_ACCESS_KEY'),
-//            'region' => env('AWS_DEFAULT_REGION'),
-//            'bucket' => env('AWS_BUCKET'),
-//            'url' => env('AWS_URL'),
-//        ],
-
-        // https://github.com/Superbalist/laravel-google-cloud-storage
-        'gcs-public-static' => [
-            'driver' => 'gcs',
-            'project_id' => env('GOOGLE_CLOUD_PROJECT_ID', 'your-project-id'),
-            'key_file' => env('GOOGLE_CLOUD_STORAGE_KEY_FILE', null), // optional: /path/to/service-account.json
-            'bucket' => env('GOOGLE_CLOUD_STORAGE_BUCKET', 'your-bucket'),
-            'path_prefix' => env('GOOGLE_CLOUD_STORAGE_PATH_PREFIX', null), // optional: /default/path/to/apply/in/bucket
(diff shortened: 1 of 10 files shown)


m90:
This still requires a chart update to provide the credentials and configuration values.

m90:
Chart update here https://github.com/wbstack/charts/pull/136

config/filesystems.php
```diff
@@ -55,33 +55,18 @@
...
-            'visibility' => 'public', // optional: public|private
+        's3-static' => [
+            'driver' => 's3',
+            'key' => env('STATIC_STORAGE_ACCESS_KEY'),
+            'secret' => env('STATIC_STORAGE_SECRET_KEY'),
+            'region' => env('STATIC_STORAGE_REGION', 'us-east-1'),
+            'bucket' => env('STATIC_STORAGE_BUCKET_NAME'),
+            'url' => env('STATIC_STORAGE_URL'),
+            'endpoint' => env('STATIC_STORAGE_ENDPOINT'),
+            'bucket_endpoint' => boolval(env('STATIC_STORAGE_USE_BUCKET_ENDPOINT', '0')),
```
m90, line 66: I deliberately went for USE in the env var name as otherwise it sounds a lot like it wants to be passed an URL. Like this the odd naming choice in the driver config is hidden from us.

m90: If there's a nicer way to get a boolean from a string in PHP please lmk.

config/filesystems.php
```diff
@@ -55,33 +55,18 @@
...
-        // https://github.com/Superbalist/laravel-google-cloud-storage
-        'gcs-public-static' => [
-            'driver' => 'gcs',
-            'project_id' => env('GOOGLE_CLOUD_PROJECT_ID', 'your-project-id'),
-            'key_file' => env('GOOGLE_CLOUD_STORAGE_KEY_FILE', null), // optional: /path/to/service-account.json
-            'bucket' => env('GOOGLE_CLOUD_STORAGE_BUCKET', 'your-bucket'),
-            'path_prefix' => env('GOOGLE_CLOUD_STORAGE_PATH_PREFIX', null), // optional: /default/path/to/apply/in/bucket
-            'storage_api_uri' => env('GOOGLE_CLOUD_STORAGE_API_URI', null), // see: Public URLs below
-            'visibility' => 'public', // optional: public|private
+        's3-static' => [
```
m90, line 58: Maybe having `s3` in the name here is a bad idea and this should rather be `static-assets` or sth?

config/filesystems.php
deer-wmde: I think the naming is useful and also I think that is a good way "to bool" :+1:

config/filesystems.php
m90: Yeah let's name it `static-assets`, I like that better as well.

PR #721, opened 2024-01-15
fix: add pessimistic lock when fetching next qs batch and marking failed

m90:
Ticket https://phabricator.wikimedia.org/T354797

### DIFF
--- a/app/Jobs/CreateQueryserviceBatchesJob.php
+++ b/app/Jobs/CreateQueryserviceBatchesJob.php
@@ -72,6 +72,12 @@ class CreateQueryserviceBatchesJob extends Job
 
     private function tryToAppendEntitesToExistingBatches(array $entityIdsFromEvents, int $wikiId): bool
     {
+        if (DB::transactionLevel() < 1) {
+            throw new \Exception(
+                'This method can only be run within the context of a transaction.'
+            );
+        }
+
         $notDoneBatches = QsBatch::where([
             ['done', '=', 0],
             ['pending_since', '=', null],
--- a/app/Jobs/CreateQueryserviceBatchesJob.php
+++ b/app/Jobs/CreateQueryserviceBatchesJob.php
@@ -77,7 +77,9 @@ class CreateQueryserviceBatchesJob extends Job
             ['pending_since', '=', null],
             ['failed', '=', false],
             ['wiki_id', '=', $wikiId],
-        ])->get();
+        ])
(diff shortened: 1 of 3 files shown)


app/Jobs/CreateQueryserviceBatchesJob.php
```diff
@@ -77,7 +77,9 @@ private function tryToAppendEntitesToExistingBatches(array $entityIdsFromEvents,
             ['pending_since', '=', null],
             ['failed', '=', false],
             ['wiki_id', '=', $wikiId],
-        ])->get();
+        ])
+            ->lockForUpdate()
```
m90, line 81: If we'd be very pedantic we could now wrap this method in sth that checks whether the code is actually run in a transaction, and if nor running in a transaction, create one for us.

Then again this is a private method and the caller is creating the transaction for us, so I would probably not do it for avoid that overengineering stench.

m90: Or a slightly easier way would be having this method check whether we're in a transaction, and throw an Exception in case we're not. Should prevent callers from accidentally using this elswhere.

deer-wmde: I agree with the whole notion, the use of the method is pretty confined so I think the danger of using it outside of an transaction context is not too high, but I also like the foresight and the latter suggestion definitely sounds like a useful guardrail

deer-wmde:
Generally this looks great to me, I'm just still lacking a bit database knowledge and am curious where the unlocking does happen again, is it something Laravel takes care of in its connection teardown or something or is it even the DB itself that does it? Just asking out of interest and trying to see if there's footgun potential (but I don't think so actually)

deer-wmde (approved):

m90:
> 
Generally this looks great to me, I'm just still lacking a bit database knowledge and am curious where the unlocking does happen again, is it something Laravel takes care of in its connection teardown or something or is it even the DB itself that does it? Just asking out of interest and trying to see if there's footgun potential (but I don't think so actually)

From what I understand the `END TRANSACTION` statement will unlock everything that was locked within the transaction.

app/Jobs/CreateQueryserviceBatchesJob.php
m90: Then let's just have it throw an exception when there is no ongoing transaction.

PR #722, opened 2024-01-16
fix(wiki-settings): include questy settings in details

m90:
Ticket https://phabricator.wikimedia.org/T354954

### DIFF
--- a/app/Wiki.php
+++ b/app/Wiki.php
@@ -138,6 +138,8 @@ class Wiki extends Model
             'wwWikibaseStringLengthMultilang',
             'wikibaseFedPropsEnable',
             'wikibaseManifestEquivEntities',
+            'wwUseQuestyCaptcha',
+            'wwCaptchaQuestions',
         ]
     );
     }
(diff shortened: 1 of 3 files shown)


app/Wiki.php
```diff
@@ -138,6 +138,8 @@ public function publicSettings()
             'wwWikibaseStringLengthMultilang',
             'wikibaseFedPropsEnable',
             'wikibaseManifestEquivEntities',
+            'wwUseQuestyCaptcha',
+            'wwCaptchaQuestions',
```
m90, line 142: I wonder if we should make this a blocklist instead? Else we'll be bitten by this again I surmise.

tarrow: Yeah, I guess things like wgSecret and the db credentials are things we would want to not expose to the manager but this list is fairly short.

m90: Today I learned there are DB credentials in `wikiSettings`. Then maybe let's keep it like this.

tarrow (approved):
cool; let's go ahead with this. I also double checked if we might be leaking something via PublicWikiResource or something but it appears not.

PR #725, opened 2024-01-18
fix: specify file visibility for public assets

m90:
Ticket https://phabricator.wikimedia.org/T354286

Note that the repro in the ticket is probably caused by some other condition.

### DIFF
--- a/app/Jobs/SetWikiLogo.php
+++ b/app/Jobs/SetWikiLogo.php
@@ -66,7 +66,7 @@ class SetWikiLogo extends Job
         $logosDir = Wiki::getLogosDirectory($wiki->id);
 
         // Upload the local image to the cloud storage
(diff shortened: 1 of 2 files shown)


app/Jobs/SetWikiLogo.php
```diff
@@ -66,7 +66,7 @@ public function handle(): void
         $logosDir = Wiki::getLogosDirectory($wiki->id);
 
         // Upload the local image to the cloud storage
-        $storage->putFileAs($logosDir, new File($this->logoPath), "raw.png");
+        $storage->putFileAs($logosDir, new File($this->logoPath), "raw.png", ['visibilty' => 'public']);
```
m90, line 69: We could also set the visibility on global level, i.e. in the driver's configuration. Not sure which approach is preferrable?

m90:
Superseded by #725

PR #727, opened 2024-01-18
Use S3 compatible driver for interacting with GCS

m90:
Ticket https://phabricator.wikimedia.org/T354744

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # api
 
+## 8x.35.1 - 22 January 2024
+- Use S3 driver for accessing static files in GCP buckets, using proper ACLs
+
 ## 8x.35.0 - 19 January 2024
 - Use pessimistic DB locks during Query Service Batch transactions
 
@@ -7,7 +10,7 @@
 - Add missing QuestyCaptcha settings to wiki details response
 - Revert "Use S3 driver for accessing static files in GCP buckets"
 
-## 8x.34.0 - 17 January 2024
+## 8x.34.0 - 17 January
 - Use S3 driver for accessing static files in GCP buckets
 
 ## 8x.33.0 - 03 January 2024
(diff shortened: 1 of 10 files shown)


deer-wmde (approved):
I'm missing some context but to me this looks like it's re-introducing the changes from https://github.com/wbstack/api/pull/719/ + ACL fixes for the problems we encountered, so I approve

CHANGELOG.md
```diff
@@ -1,10 +1,13 @@
 # api
 
+## 8x.34.2 - 18 January
+- Use S3 driver for accessing static files in GCP buckets, using proper ACLs
```
tarrow, line 4: I wanted to confirm I understand this correctly:

This applies exactly the same logic as #719 but also adds the `'visibility' => 'public'` setting to all places we save files. For what it's worth I'd have no objection to also setting this for the whole `Storage disk` object since I can see us forgetting to set this in future and confusing us.

Indeed if I understood our possible desire to move towards Unified Bucket access (e.g. with no per object ACLs) perhaps we don't want to be having the s3 driver setting these things at at; we might want to advance to the 3.0 version (perhaps with a small feature outage while we do it). This would then allow Laravel to return the the state it was in before of "totally unaware of anything about the file visibility of uploaded logos".

These changes alone are not sufficient to make the feature work (i.e. logos won't be visible to the public) though since now the service account this is using needs to also have the right to make the ACL changes (since 1.0 of league/flysystem-aws-s3-v3) dies if it can't make the ACL changes it tries to.

m90: > These changes alone are not sufficient to make the feature work (i.e. logos won't be visible to the public) though since now the service account this is using needs to also have the right to make the ACL changes (since 1.0 of league/flysystem-aws-s3-v3) dies if it can't make the ACL changes it tries to.

I'm not entirely sure what this paragraph is trying to say.

PR #730, opened 2024-01-25
Laravel 10 Upgrade

deer-wmde:
https://phabricator.wikimedia.org/T341797

WIP


app/Jobs/CirrusSearch/ForceSearchIndex.php
```diff
@@ -11,7 +11,7 @@
  * 
  * Example:
  * 
- * php artisan job:dispatchNow CirrusSearch\\ForceSearchIndex id 1 0 1000
+ * php artisan job:dispatchSync CirrusSearch\\ForceSearchIndex id 1 0 1000
```
m90, line 14: I wonder if we want `dispatch` instead. If I understand it correctly `dispatchSync` will not use the existing queue setup, but somehow bypass this. Right now, I would think we want this to run together with the other jobs instead.

But maybe this is also something we should look into with https://phabricator.wikimedia.org/T342866

app/User.php
```diff
@@ -76,6 +76,16 @@ class User extends Authenticatable implements MustVerifyEmail
         'card_last_four',
     ];
 
+    /**
+     * The attributes that should be cast.
+     *
+     * @var array<string, string>
+     */
+    protected $casts = [
+        'email_verified_at' => 'datetime',
```
m90, line 85: Is this a drive by improvement or something we _have_ to when using Laravel 10?

config/mail.php
```diff
@@ -28,14 +28,15 @@
...
-    |            "postmark", "log", "array"
+    | Supported: "smtp", "sendmail", "mailgun", "ses", "ses-v2",
+    |            "postmark", "log", "array", "failover", "roundrobin"
     |
     */
 
     'mailers' => [
         'smtp' => [
             'transport' => 'smtp',
+            'url' => env('MAIL_URL'),
```
m90, line 39: Are we ever setting this?

phpunit.xml.bak
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
m90, line 1: Do we need this file or is it a leftover?

m90 (commented):
This looks pretty solid already. I left a few questions inline, nothing blocking.

app/Jobs/CirrusSearch/ForceSearchIndex.php
deer-wmde: I just realized this console command is completely separate from the framework anyway... https://github.com/mxl/laravel-job/blob/master/src/Commands/DispatchNow.php

But still I think you are right, `dispatch` is probably a better choice most of the time. I'll change the comments accordingly

deer-wmde: There are some more critical instances where I'd leave it though, MediawikiUpdate.php for example https://github.com/wbstack/api/blob/e59eda7de24f1baccbb7a02c285f684eebe93ecd/app/Jobs/MediawikiUpdate.php

config/mail.php
deer-wmde: I don't think so, probably best to not transfer this change

PR #731, opened 2024-01-26
Replace `laravel-stackdriver` package

deer-wmde:
https://phabricator.wikimedia.org/T344377

- remove `gluedev/laravel-stackdriver`
- add `absszero/laravel-stackdriver-error-reporting`


app/Exceptions/Handler.php
```diff
@@ -36,14 +35,12 @@ class Handler extends ExceptionHandler
...
 
     /**
-     * Report or log an exception.
-     *
-     * @param  \Throwable  $e
-     * @return void
+     * Register the exception handling callbacks for the application.
      */
-    public function report(Throwable $e)
+    public function register()
```
tarrow, line 40: We talked about this; we just wanted to check that this will behave in the same way as before but is the new recommended way to do this.

tarrow (commented):
Assuming this one comment is "fine" and that the error reports coming out look the same then I'm happy for this to me merged :)

deer-wmde:
This is an example of the reported errors that I triggered on monday on staging for testing this:
[GCP Error reporting](https://console.cloud.google.com/errors/detail/CObm2ePvpY_ygAE;service=WBaaS%20Wikibase%20Dev;version=1.0.0;time=P30D?project=wikibase-cloud)
[GCP Logs](https://console.cloud.google.com/logs/query;query=error_groups.id%3D%22CObm2ePvpY_ygAE%22%0AlogName:%22app-error%22%0Aresource.type%3D%22global%22%0Aresource.labels.project_id%3D%22wikibase-cloud%22;cursorTimestamp=2024-01-29T17:15:44.704252Z;startTime=2024-01-29T16:46:14.704Z;endTime=2024-01-29T17:46:14.704Z?project=wikibase-cloud)

Unfortunately it seems there is no matching naturally ocurring instance of this exact error in the last 30 days but what we still can see here is that it gets reported just as other errors before and it also turns up in the console logs

app/Exceptions/Handler.php
deer-wmde: Looking at the [docs](https://laravel.com/docs/10.x/errors) this seems to be recommended, although I now realized a reason to override the report function may be if you want to call it yourself when there is no Exception (see section [The report helper](https://laravel.com/docs/10.x/errors#the-report-helper))

deer-wmde: I don't think we do this anywhere though

deer-wmde: Okay I have to correct myself again, after testing locally I could verify that the handler still gets called this way, so we should be good.

config/error_reporting.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+return [
+    // Enable error reporting
+    'enable' => (bool) env('STACKDRIVER_ENABLED', false),
+    'logName' => env('STACKDRIVER_LOGGING_NAME', 'error-log'),
+    'serviceId' => env('STACKDRIVER_ERROR_REPORTING_SERVICE_ID', env('APP_NAME')),
+    'versionId' => env('STACKDRIVER_ERROR_REPORTING_VERSION_ID', '1.0.0'),
+    'LoggingClient' => [
```
tarrow, line 9: As we mentioned in person it could be good to take 5mins to check if there is a way to make this logging appears as though its from a k8s container and add all the appropriate metadata to make it good; if not then no worries!

tarrow (approved):

config/error_reporting.php
tarrow: Cool, let's just go ahead and merge this so we're unblocked on the laravel 10 update. I think we should then (not immediately) revisit our whole strategy and consider "just" logging to stdout/stderr similar to mediawiki

deer-wmde: > I think we should then (not immediately) revisit our whole strategy and consider "just" logging to stdout/stderr similar to mediawiki

sounds interesting, happy to talk about and improve this

PR #736, opened 2024-02-07
fix(qs-batches): retry running transactions that require db locks

m90:
This is a fixup for #721 

Apparently transactions that cannot acquire a lock now fail (which is expected), so we should retry them afterwards a couple of times before giving up. Laravel's `DB::transaction` can do this for us.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # api
 
+## 9x.0.2 - 07 February 2024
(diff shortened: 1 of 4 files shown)


m90:
> I don't think this retry behaviour immediately needs to be added there

Yup, I would think we only need to retry when we actually lock. Else, the failure is very unlikely to succeed on second attempt.

PR #737, opened 2024-02-08
Release 9x.0.3

tarrow:
Updates a number of github action dependencies.

Highly unlikely to be an issue but by releasing and deploying a version we'll not have these changes confused with other work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # api
 
+## 9x.0.3 - 08 February 2024
+- Updates a number of github action dependencies
+
 ## 9x.0.2 - 07 February 2024
 - Retry DB transactions that lock
 


m90:
We had this discussion before, but do we need to release something that is affecting Actions only?

tarrow:
We don't need to; tbh if deniz is about to do a release this will catch bugs in these actions changes anyway. If there's a chance these sit merged for 2 weeks though I think it's worth it. Otherwise a bug might appear and it's a pain to discover what it was from

deer-wmde:
closed, included here https://github.com/wbstack/api/commit/f0304aef9a884be015d71dcb73b5cb1bbc450ffc

m90:
> this will catch bugs in these actions changes anyway

Did this ever really happen? My subjective take is that my time is better spent preventing the introduction of grave bugs I ship myself instead of looking for things I don't even know how they would look like.

I guess Argo will free us from this anyways?

m90:
Thinking about this more: if we are bothered by the idea of our build setup possibly not giving us reproducible builds from the same source, we should rather look into how to get reproducible builds, which would be time well spent.

PR #753, opened 2024-03-11
Use Laravel Horizon as a queue supervisor

m90:
Tickets https://phabricator.wikimedia.org/T342866

Docs for the horizon package are here: https://laravel.com/docs/10.x/horizon


.github/workflows/composer.test.yml
```diff
@@ -24,6 +24,7 @@ jobs:
       with:
         php_version: 8.2
         command: install
+        args: --ignore-platform-req=ext-pcntl
```
m90, line 27: I do lack the PHP knowledge to be able to tell whether this is ok to do. Things continued to work for me locally :tm:, but maybe that's a happy coincidence?

Dockerfile
```diff
@@ -4,21 +4,24 @@ COPY ./composer.json /tmp/src1/composer.json
...
 
 WORKDIR /tmp/src1
-RUN composer install --no-dev --no-progress --no-autoloader
+RUN composer install --no-dev --no-progress --no-autoloader --ignore-platform-req=ext-pcntl
 
 COPY ./ /tmp/src2
 RUN cp -r /tmp/src1/vendor /tmp/src2/vendor
 WORKDIR /tmp/src2
-RUN composer install --no-dev --no-progress --optimize-autoloader
+RUN composer install --no-dev --no-progress --optimize-autoloader --ignore-platform-req=ext-pcntl
```
m90, line 12: I'm not entirely sure why this Dockerfile installs dependencies (twice?) in a dedicated image before copying dependencies over (not having to install composer in the production image?).

We could get rid of the extra flag if we installed it in the next step, but maybe it's ok as it is.

start.sh
```diff
@@ -15,6 +14,10 @@ if [ "$role" = "app" ]; then
 
     exec apache2-foreground
 
+elif [ "$role" = "queue" ] && [ "$HORIZON_ENABLED" = "1" ]; then
+
+    php /var/www/html/artisan horizon
```
tarrow, line 19: The [deploying horizon docs](https://laravel.com/docs/10.x/horizon#deploying-horizon) naturally talk about a more metal and less k8s/containerized environment.

I wonder if we need to do anything special here (for example they use `supervisor`). For the case of this terminating I think that it won't actually be necessary because I would expect the container in the pod to restart. Is that right?

I was also unclear if we need to handle some special cases when we are pushing new images. Is the code shared between pods in some way that necessitates us running this `php artisan horizon:terminate`? Or will creating the new pods with the new image be sufficient? Again I assume yes?

This is a long way to say that I expect us to need to do nothing and this is great but I wanted to ask.

app/Providers/HorizonServiceProvider.php
```diff
@@ -0,0 +1,36 @@
...
+
+        // Horizon::routeSmsNotificationsTo('15556667777');
+        // Horizon::routeMailNotificationsTo('example@example.com');
+        // Horizon::routeSlackNotificationsTo('slack-webhook-url', '#channel');
+    }
+
+    /**
+     * Register the Horizon gate.
+     *
+     * This gate determines who can access Horizon in non-local environments.
```
tarrow, line 26: What is a non-local environment? I notice that we can see the shiny UI at http://api.wbaas.localhost/horizon/ in the local environment. I'm kinda assuming we don't want this in production but I'm unclear if it will be present there.

I don't understand how horizon or laravel would know minikube is local but GKE isn't. I'm imagining they look very similar from the perspective of the pod

m90: Laravel sets an `environment` which for us is either `local`, `staging` or `production`. If I understand this correctly, we should serve the Dashboard in all envs, but there is no way to authenticate when deployed to staging or production.

start.sh
m90: I would think :tm: all of the "supervisor" duties explained in the documentation are meaningless for us as we put the thing in a container and let k8s do the supervision for us.

> Or will creating the new pods with the new image be sufficient? Again I assume yes?

I would assume yes as well. I haven't noticed anything odd when rolling out new changes in local.

tarrow (approved):

PR #761, opened 2024-03-14
fix(horizon): add stub configs for all environments that are deployed

m90:
It seems the default configuration only declares supervisors for `local` and `production` envs (https://github.com/laravel/horizon/blob/0f6a1dd80f7ec3f448a46578c07f0a872ebb07a9/config/horizon.php#L198-L212), which made #753 not do anything when deployed to staging.

This PR fixes this by declaring stubs for all environments known to wikibase.cloud.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # api
 
+## 10x.1.1 - 14 March 2024
+- Declare Horizon configuration stubs for all known environments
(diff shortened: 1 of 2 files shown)


m90:
> no issue, I assume, with overwriting/blanking the maxProcesses etc. since our environment variables overwrite these?

Exactly.

PR #771, opened 2024-03-25
fix: horizon supervisor does not consider job level timeouts

m90:
Ticket https://phabricator.wikimedia.org/T342866

It seems that when using Horizon to supervise jobs, job level timeouts are not being considered. Instead only the Horizon config itself will be used. Jobs can still timeout themselves.

Currently, this means that long running jobs like `PollForMediaWikiJobs` will timeout after 60s in production.

As we probably still want job level granularity for setting this value, this PR changes:
- Use a timeout of 1h at Horizon level. This will catch jobs that are left hanging when they really shouldn't be
- Jobs themselves now define a default timeout of 60s (as previously), and jobs that need to run longer need to explicitly declare a `$timeout` property


app/Jobs/PollForMediaWikiJobsJob.php
```diff
@@ -10,7 +10,7 @@
 
 class PollForMediaWikiJobsJob extends Job implements ShouldQueue, ShouldBeUnique
 {
-    public $timeout = 3600;
+    public $timeout = 1800;
```
m90, line 13: Currently this takes something around 3 to 7 minutes in production (depending on current load), so I think it's fine to lower this while we're at it.

deer-wmde (approved):
Looks sensible to me

PR #774, opened 2024-03-27
chore(jobs): default to all clusters in Elasticsearch init job

AndrewKostka:


tarrow (approved):
gotta be deployed in sync with https://github.com/wbstack/mediawiki/pull/437

app/Jobs/CirrusSearch/ElasticSearchIndexInit.php
```diff
@@ -6,6 +6,13 @@
 
 class ElasticSearchIndexInit extends CirrusSearchJob
 {
+    private $cluster;
+
+    public function __construct( int $wikiId, string $cluster = 'all' ) {
```
m90, line 11: Question: At which point will someone pass a different value than `all` here? Asking as I wonder why we don't just hardcode it in `getQueryParams`?

PR #778, opened 2024-04-04
Create a model for Entities count

dati18:


app/Jobs/UpdateWikiEntitiesCountJob.php
```diff
@@ -0,0 +1,93 @@
...
+    {
+        $this->apiUrl = getenv('PLATFORM_MW_BACKEND_HOST').'/w/api.php';
+        $allWiki = Wiki::all();
+        foreach ($allWiki as $wiki) {
+            try{
+                $this->updateEntitiesCount($wiki);
+            } catch (\Exception $ex) {
+                $this->job->markAsFailed();
+                Log::error(
+                    'Failure polling wiki '.$wiki->getAttribute('domain').' for sitestats: '.$ex->getMessage()
```
m90, line 26:
```suggestion
                    'Failure polling wiki '.$wiki->getAttribute('domain').' for item and property counts: '.$ex->getMessage()
```

app/Jobs/UpdateWikiEntitiesCountJob.php
```diff
@@ -0,0 +1,93 @@
...
+
+class UpdateWikiEntitiesCountJob extends Job implements ShouldBeUnique
+{
+    private const NAMESPACE_ITEM = 122;
+    private const NAMESPACE_PROPERTY = 120;
+    protected string $apiUrl;
+
+    public function handle (): void
+    {
+        $this->apiUrl = getenv('PLATFORM_MW_BACKEND_HOST').'/w/api.php';
```
m90, line 18: Nit: this is probably better done in the class constructor as it never changes.

app/WikiEntitiesCount.php
```diff
@@ -0,0 +1,26 @@
...
+        'items_count',
+        'properties_count',
+        'wiki_id'
+    ];
+
+    protected $fillable = self::FIELDS;
+
+    protected $visible = self::FIELDS;
+
+    protected $casts = [
```
m90, line 22: Question: seeing these fields are declared as integers in the database schema already, is the cast still needed?

tests/Jobs/UpdateWikiEntitiesCountJobTest.php
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    public function testEmpty()
+    {
+        Bus::fake();
+        Http::fake();
+
+        Bus::assertNothingDispatched();
+        Http::assertNothingSent();
+    }
```
m90, line 33: Issue: This test currently doesn't test the job as its `handle` method is never called. See the test below to check how a mock job is being created and its handle method invoked.

tests/Jobs/UpdateWikiEntitiesCountJobTest.php
```diff
@@ -0,0 +1,191 @@
...
+use Illuminate\Support\Facades\Http;
+use PHPUnit\Framework\MockObject\Exception;
+use Tests\TestCase;
+
+class UpdateWikiEntitiesCountJobTest extends TestCase
+{
+    use RefreshDatabase;
+
+    public function setUp(): void {
+        parent::setUp();
```
m90, line 19: Nitpick: If there is no setup that is specific to `UpdateWikiEntitiesCountJobTest` to be done, there is no need to redeclare `setUp`  like this. Same goes for `tearDown`

m90 (commented):
The code for the job that fetches the counts looks mostly solid to me, but before I dive into too much detail,  I'd have an architectural question about this:

What is the reason you decided to create a setup that works like:

1. create a job for fetching counts and persisting them for each wiki
2. run this job once a day
3. in the platform stats summary job, which is also running daily, read these values and include them in the stats object that is being read by the dashboard

instead of:

1. fetch the counts while creating the platform summary stats and include them in the stats object without persisting anything

Is the level of indirection introduced by persisting the data in the database needed as we are planning to use the data somewhere else?

dati18:
> The code for the job that fetches the counts looks mostly solid to me, but before I dive into too much detail, I'd have an architectural question about this:
> 
> What is the reason you decided to create a setup that works like:
> 
> 1. create a job for fetching counts and persisting them for each wiki
> 2. run this job once a day
> 3. in the platform stats summary job, which is also running daily, read these values and include them in the stats object that is being read by the dashboard
> 
> instead of:
> 
> 1. fetch the counts while creating the platform summary stats and include them in the stats object without persisting anything
> 
> Is the level of indirection introduced by persisting the data in the database needed as we are planning to use the data somewhere else?

I thought that maybe we need items and props count of individual wiki in some case, and it can be also modified to add more stats (i.e Lexeme for example). I'm aware that it can hurt the performance and redundant somehow.

m90:
> I thought that maybe we need items and props count of individual wiki in some case, and it can be also modified to add more stats (i.e Lexeme for example)

I would think unless we know we need it, we assume [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) and skip the need for persisting it until someone else actually wants to read it.

> I'm aware that it can hurt the performance and redundant somehow.

I think performance should be ok in this case, but it would definitely create a lag in the freshness of data being displayed on the dashboard.

The only pro I can see for handling this in a dedicated job is that incorporating this in the platform stats summary job will make the job quite a bit slower, but since we now can safely run multiple jobs in parallel (as per #753) I would also think this is ok until we find out it's causing problems.

dati18:
> persist



> > I thought that maybe we need items and props count of individual wiki in some case, and it can be also modified to add more stats (i.e Lexeme for example)
> 
> I would think unless we know we need it, we assume [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) and skip the need for persisting it until someone else actually wants to read it.
> 
> > I'm aware that it can hurt the performance and redundant somehow.
> 
> I think performance should be ok in this case, but it would definitely create a lag in the freshness of data being displayed on the dashboard.
> 
> The only pro I can see for handling this in a dedicated job is that incorporating this in the platform stats summary job will make the job quite a bit slower, but since we now can safely run multiple jobs in parallel (as per #753) I would also think this is ok until we find out it's causing problems.

yes I got your point, I also notice that `UpdateWikiEntitiesCountJob` run after `UpdateWikiSiteStatsJob`. 

So what should I do next? 1. Change the Job schedule to run early (pro: quick fix, con: architect is still the same) or 2. Scrape off the Model and modify PlatformSummaryStatsJob to fetch the count?

dati18:
> > I thought that maybe we need items and props count of individual wiki in some case, and it can be also modified to add more stats (i.e Lexeme for example)
> 
> I would think unless we know we need it, we assume [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) and skip the need for persisting it until someone else actually wants to read it.
> 
> > I'm aware that it can hurt the performance and redundant somehow.
> 
> I think performance should be ok in this case, but it would definitely create a lag in the freshness of data being displayed on the dashboard.
> 
> The only pro I can see for handling this in a dedicated job is that incorporating this in the platform stats summary job will make the job quite a bit slower, but since we now can safely run multiple jobs in parallel (as per #753) I would also think this is ok until we find out it's causing problems.

And I also think we should as Product and Design if they need these EntitiesCOunt stats (like for the Discovery page)

m90:
> So what should I do next?

I would suggest we

1. Move the existing code for fetching the entity counts into the PlatformSummaryStatsJob
2. Instead of reading the newly introduced stats from the DB, call the method from 1. that fetches them on the spot
3. Adjust the tests for the PlatformSummaryStatsJob
4. Remove unneeded code for persisting the model

Does that make sense? Even if it feels big I would think you can reuse what you have already written and just rearrange it.

> And I also think we should as Product and Design if they need these EntitiesCOunt stats

Right now, we don't. If we build an application that caters for everything that could happen at some point, it'll be pretty complex. If we really do need this at some point, we can still adjust the simple solution we have then.

dati18:
> > So what should I do next?
> 
> I would suggest we
> 
> 1. Move the existing code for fetching the entity counts into the PlatformSummaryStatsJob
> 2. Instead of reading the newly introduced stats from the DB, call the method from 1. that fetches them on the spot
> 3. Adjust the tests for the PlatformSummaryStatsJob
> 4. Remove unneeded code for persisting the model
> 
> Does that make sense? Even if it feels big I would think you can reuse what you have already written and just rearrange it.
> 
> > And I also think we should as Product and Design if they need these EntitiesCOunt stats
> 
> Right now, we don't. If we build an application that caters for everything that could happen at some point, it'll be pretty complex. If we really do need this at some point, we can still adjust the simple solution we have then.

That is not too complex for me, but I will create a new branch and PR for it rather continue working on this branch because it would be a mess.

dati18:
> > So what should I do next?
> 
> I would suggest we
> 
> 1. Move the existing code for fetching the entity counts into the PlatformSummaryStatsJob
> 2. Instead of reading the newly introduced stats from the DB, call the method from 1. that fetches them on the spot
> 3. Adjust the tests for the PlatformSummaryStatsJob
> 4. Remove unneeded code for persisting the model
> 
> Does that make sense? Even if it feels big I would think you can reuse what you have already written and just rearrange it.
> 
> > And I also think we should as Product and Design if they need these EntitiesCOunt stats
> 
> Right now, we don't. If we build an application that caters for everything that could happen at some point, it'll be pretty complex. If we really do need this at some point, we can still adjust the simple solution we have then.

I will archive this branch, you know, just in case

dati18:
> @dati18 I'm sorry if this ends up feeling frustrating; I certainly didn't raise any of these objections in our earlier discussions and was partly convincing you that:
> 
> a) there is a certain nice modularity to splitting the the jobs apart b) we may actually want individual entity stats for each wiki (indeed perhaps going on to add entities
> 
> Hopefully, like @m90 says, actually hopping between these two designs is pretty fast.

I'm working on it. :) no worry because it makes sense

m90:
> there is a certain nice modularity to splitting the the jobs apart

@tarrow While I see that argument, I think we don't win anything if we do this in a piece-meal fashion. We could think about a follow up task that refactors the PlatformSummaryStatsJob into a parent job that fans out to multiple jobs that calculate / collect metrics, and then emits the metrics when the child jobs are done.

However, if only one portions of the metrics is collected like this and everything else is living in the top level job, I would think this is more confusing than helpful and also makes debugging / error handling very hard. Maybe we can talk about such an idea in a refinement at some point.

m90:
This was done elsewhere, closing therefore.

PR #785, opened 2024-04-16
fix(site-stats): enforce uniqueness of wiki_site_stats per wiki, lock table when updating

m90:
Ticket https://phabricator.wikimedia.org/T362667

### DIFF
--- a/database/migrations/2024_04_16_165700_enforce_site_stats_constraint.php
+++ b/database/migrations/2024_04_16_165700_enforce_site_stats_constraint.php
@@ -25,8 +25,13 @@ class EnforceSiteStatsConstraint extends Migration
      */
     public function down()
     {
+        // foreign key constraints need to be disabled as per https://github.com/laravel/framework/issues/13873
+        Schema::disableForeignKeyConstraints();
         Schema::table('wiki_site_stats', function (Blueprint $table) {
-            $table->dropUnique('wiki_id');
+            // The column name HAS to be wrapped in an array so Laravel can
+            // figure out the relation name.
+            $table->dropUnique(['wiki_id']);
         });
+        Schema::enableForeignKeyConstraints();
     }
 }


app/Jobs/UpdateWikiSiteStatsJob.php
```diff
@@ -74,7 +75,8 @@ private function updateSiteStats (Wiki $wiki): void
                 $update[$field] = $value;
             }
         }
-
-        $wiki->wikiSiteStats()->updateOrCreate(['wiki_id' => $wiki->id], $update);
+        DB::transaction(function () use ($wiki, $update) {
+            $wiki->wikiSiteStats()->lockForUpdate()->updateOrCreate(['wiki_id' => $wiki->id], $update);
```
tarrow, line 79: praise: Looks good; I read up a little on lockForUpdate and it look like it would do what we need it to

m90:
Migrations apply fine locally:

```
root@api-queue-default-6ff6649d4c-9hxxl:/var/www/html# php artisan migrate

   INFO  Running migrations.

  2024_04_16_165700_enforce_site_stats_constraint ......................................................................................... 6ms DONE

root@api-queue-default-6ff6649d4c-9hxxl:/var/www/html# php artisan migrate:rollback --step 1

   INFO  Rolling back migrations.

  2024_04_16_165700_enforce_site_stats_constraint ......................................................................................... 7ms DONE

root@api-queue-default-6ff6649d4c-9hxxl:/var/www/html# php artisan migrate

   INFO  Running migrations.

  2024_04_16_165700_enforce_site_stats_constraint ......................................................................................... 6ms DONE
```

tarrow (approved):
looks good to me; I guess "don't forget to run the migrations" after deploying this. I wonder if we really ought to trigger these with some helm hook or something smart on future.

m90:
> I wonder if we really ought to trigger these with some helm hook or something smart on future.

That would be fantastic.

PR #792, opened 2024-05-06
Allow auth token in cookies

m90:
Ticket https://phabricator.wikimedia.org/T363100


tarrow (approved):
I have no more comments after our discussion last week :)

PR #794, opened 2024-05-08
fix: platform stats summary job handling inactive wikis

tarrow:
It appears since we created this job that we didn't properly handle the situation where there most recent revision isn't found for some reason. In this case we should assume that it hasn't been edited recently.

This could have happened since we updated mediawiki causing a change in the behaviour of the revision table or it could always have been non-functional

We should follow up with not relying on directly querying the mediawiki db for these stats and trying to find a more stable public api for it.

Bug: T364452

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # api
 
+## 10x.5.0 - 8 May 2024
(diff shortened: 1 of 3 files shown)


tarrow:
Roger that

PR #795, opened 2024-05-08
Add items and properties counts to PlatformStatsSummaryJob

dati18:


app/Jobs/PlatformStatsSummaryJob.php
```diff
@@ -163,7 +181,54 @@ public function handle( DatabaseManager $manager ): void
         if( !App::runningUnitTests() ) {
             print( json_encode($summary) . PHP_EOL );
         }
+    }
+
+    private function fetchPagesInNamespace(string $wikiDomain, int $namespace): array
```
m90, line 186: If I understand it correctly, this method is copied verbatim from the `RebuildQueryserviceData` command. To remove the duplication, what do you think about doing the following:

- create a `PageFetcher` trait in `App\Traits` that contains this method as well as the `apiUrl` property.
- `use` this trait here as well as in `RebuildQueryserviceData`

m90 (commented):
This looks good, I left an idea about how to get rid of the duplication of code that this now introduces. Let me know what you think.

app/Jobs/PlatformStatsSummaryJob.php
dati18: done

tests/Jobs/PlatformStatsSummaryJobTest.php
```diff
@@ -98,8 +102,46 @@ public function testGroupings()
                 'version' => 'asdasdasdas',
                 'prefix' => 'asdasd',
                 'wiki_id' => $wiki->id
-            ]);  
+            ]);
+            Bus::fake();
```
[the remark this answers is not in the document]
dati18: I used it for debugging or some other reasons, I forgot to delete

tests/Jobs/PlatformStatsSummaryJobTest.php
```diff
@@ -169,9 +210,11 @@ public function testGroupings()
                 "total_non_deleted_active_users" => 1,
                 "total_non_deleted_pages" => 4,
                 "total_non_deleted_edits" => 3,
-                "platform_summary_version" => "v1"
+                "platform_summary_version" => "v1",
+                "total_items_count" => 36,
```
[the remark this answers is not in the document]
dati18: I fixed it and add a comment so people understand

app/Traits/PageFetcher.php
```diff
@@ -0,0 +1,60 @@
...
+use Illuminate\Support\Facades\Http;
+
+trait PageFetcher
+{
+    private string $apiUrl;
+
+    //this function is used to fetch pages on namespace
+    function fetchPagesInNamespace(string $wikiDomain, int $namespace): array
+    {
+        $this->apiUrl = getenv('PLATFORM_MW_BACKEND_HOST').'/w/api.php';
```
m90, line 14: Nitpick: I feel the trait reading configuration values all by itself is a bit of "overstepping" its own responsibilities (the trait shouldn't have any idea how it's being configured), so how about:

- setting the `apiUrl` prop in the class constructors as before?
- in this method, raise a `RuntimeException` in case it hasn't been set

app/Jobs/PlatformStatsSummaryJob.php
```diff
@@ -28,17 +30,25 @@
 */
 class PlatformStatsSummaryJob extends Job
```
m90, line 31: This job now probably requires an increased `timeout` value so that in production it doesn't get killed by the supervisor while still fetching data. Other jobs that poll all wikis currently use a value of `1800` so maybe it's a good idea to use the same one here. E.g.: https://github.com/wbstack/api/blob/5037c4f8ad4923966eba350c7bfa56f812616bef/app/Jobs/PollForMediaWikiJobsJob.php#L13

m90 (commented):
I left two more final comments, other than that :rocket:

app/Jobs/PlatformStatsSummaryJob.php
```diff
@@ -2,10 +2,12 @@
 
 namespace App\Jobs;
 
+use App\Traits;
 use App\Helper\MWTimestampHelper;
 use App\Wiki;
 use App\User;
 use Illuminate\Database\DatabaseManager;
+use Illuminate\Support\Facades\Http;
```
m90, line 10: If I'm not mistaken this is not used anymore now that the code has been moved to the trait, but it also won't really hurt much.

m90 (approved):
Looks good now, let's try this out in staging.

app/Jobs/PlatformStatsSummaryJob.php
dati18: oh yes I missed that

m90:
Would you mind adding a CHANGELOG entry still before merging? No need to get this reviewed though.

PR #796, opened 2024-05-13
feat: Helper class to work with MWTimestamps

tarrow:


app/Helper/MWTimestampHelper.php
```diff
@@ -0,0 +1,27 @@
...
+*/
+
+namespace App\Helper;
+use Carbon\CarbonImmutable;
+
+class MWTimestampHelper {
+
+    private const MWTimestampFormat = 'YmdHis';
+
+    static public function getCarbonFromMWTimestamp( string $MWTimestamp ): CarbonImmutable {
```
m90, line 16: I don't know too much about MediaWiki timestamps, but would it make sense to have a second optional parameter that would allow users to pass in a non-default format? Or is the default format the only one we'll ever see in the context of this application?

tarrow: I don't know much either; the first I can remember reading about them was today.

From the docs this is the the format I would expect us to encounter.

I think this is a reason that poking directly into the DB is a "bad idea". I would imagine in future we could see a migration from these to a native maria timestamp.

m90: Ok, then let's YAGNI

m90 (approved):

PR #803, opened 2024-05-16
Add user admin flag and deletion reasons to the DB

rosalieper:
https://phabricator.wikimedia.org/T347865
The admins here are WMDE wikibase cloud team

### DIFF
--- a/app/User.php
+++ b/app/User.php
@@ -32,7 +32,7 @@ use Illuminate\Database\Eloquent\Factories\HasFactory;
  * @property-read int|null $notifications_count
  * @property-read \Illuminate\Database\Eloquent\Collection|\Laravel\Passport\Token[] $tokens
  * @property-read int|null $tokens_count
- * @method static \Database\Factories\UserFactory factory(...$parameters)
+ * @method static \Database\Factories\UserFactory   factory(...$parameters)
  * @method static \Illuminate\Database\Eloquent\Builder|\App\User newModelQuery()
  * @method static \Illuminate\Database\Eloquent\Builder|\App\User newQuery()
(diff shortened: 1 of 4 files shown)


database/migrations/2024_05_15_042959_add_admin_role_to_users.php
```diff
@@ -0,0 +1,28 @@
...
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('users', function (Blueprint $table) {
+            $table->boolean('is_admin')->default(false);
```
m90, line 15: Idea: if we made this an `int`, we could enumerate "levels" of admins, i.e. `0` would be a plain user, `1` would be a "super admin" and when needed, we could start adding more values to model some sort of permissions.

m90: I have no objections against using a boolean for now though, so feel free to disregard this.

app/User.php
```diff
@@ -74,6 +74,7 @@ class User extends Authenticatable implements MustVerifyEmail
         'trial_ends_at',
         'card_brand',
         'card_last_four',
+        'is_admin',
```
tarrow, line 77: I think this should probably be in "fillable" rather than hidden. What do you think?

app/Wiki.php
```diff
@@ -52,6 +52,7 @@ class Wiki extends Model
         'domain',
         'description',
         'is_featured',
+        'wiki_deletion_reason',
```
tarrow, line 55: looks good to me

tarrow (approved):
nice, let's see this merged then :)

PR #804, opened 2024-05-16
fix(stats): namespace ids for counts are flipped

m90:
Ticket https://phabricator.wikimedia.org/T314481

We noticed the numbers for item and property count look off, and investigation shows this is because the namespace ids were wrongly mapped. This PR rectifies the mapping.

It also adds some error handling for networked actions, allowing for some calls to fail, else we'll end up with no metrics on a single network hiccup.

### DIFF
--- a/app/Jobs/PlatformStatsSummaryJob.php
+++ b/app/Jobs/PlatformStatsSummaryJob.php
@@ -38,8 +38,8 @@ class PlatformStatsSummaryJob extends Job
 
     private $platformSummaryStatsVersion = "v1";
 
-    private const NAMESPACE_ITEM = 122;
-    private const NAMESPACE_PROPERTY = 120;
+    private const NAMESPACE_ITEM = 120;
+    private const NAMESPACE_PROPERTY = 122;
 
     public function __construct() {
         $this->inactiveThreshold = Config::get('wbstack.platform_summary_inactive_threshold');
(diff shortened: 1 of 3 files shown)


app/Jobs/PlatformStatsSummaryJob.php
```diff
@@ -85,8 +85,18 @@ public function prepareStats( array $allStats, $wikis): array {
             }
 
             //add items and properties counts of the wiki to the corresponded arrays
-            array_push($itemsCount, count($this->fetchPagesInNamespace($wiki->domain, self::NAMESPACE_ITEM)));
-            array_push($propertiesCount, count($this->fetchPagesInNamespace($wiki->domain, self::NAMESPACE_PROPERTY)));
+            try {
+                $nextItemCount = count($this->fetchPagesInNamespace($wiki->domain, self::NAMESPACE_ITEM));
+                array_push($itemsCount, $nextItemCount);
+            } catch (\Exception $ex) {
+                Log::warning("Failed to fetch item count for wiki ".$wiki->domain.", will use 0 instead.");
```
m90, line 92: This is not ideal, but I still think it's better than having this job fail on a single network error or a wiki misbehaving for unknown reasons.

Maybe an idea would be to define a failure budget (maybe 5?) and bail if more than that number of calls failed?

tarrow (approved):

PR #817, opened 2024-06-19
fix: enforce unique lifecycle constraint per wiki on db level

m90:
Ticket: https://phabricator.wikimedia.org/T364991, follows up on #808 

For whatever reason we currently create multiple event objects per wiki when all of the code is written with the assumption that this is a one to one relationship.

This PR adds:
- a migration that cleans up the existing data and enforces uniqueness at db level
- acquires a lock on the table before writing new objects

### DIFF
--- a/tests/Routes/Wiki/ConversionMetricTest.php
+++ b/tests/Routes/Wiki/ConversionMetricTest.php
@@ -47,28 +47,6 @@ class ConversionMetricTest extends TestCase
         $response->assertSee('two.wikibase.cloud');
     }
 
-    public function testDownloadJsonWithDuplicateLifecycleEvents() {
-        $current_date = CarbonImmutable::now();
-        $wiki = $this->createTestWiki('one.wikibase.cloud', 10, 9, 2);
-        $wiki->wikiLifecycleEvents()->create([
-            'last_edited' => $current_date->subWeeks(1),
-            'first_edited' => $current_date->subWeeks(9)
-        ]);
-        $wiki->save();
-        $response = $this->getJson($this->route);
-        $response->assertStatus(200);
-        $response->assertJsonFragment(
-            [
-                'domain' => 'one.wikibase.cloud',
-                'time_to_engage_days' => 7,
-                'time_before_wiki_abandoned_days' => null,
-                'number_of_active_editors' => 0,
-                'first_edited_time' => $current_date->subWeeks(9),
-                'last_edited_time' => $current_date->subWeeks(1)
-            ]
-        );
-    }
-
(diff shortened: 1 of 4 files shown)


m90:
Seems there is a lot of duplicate data in the database:

```
> WikiLifecycleEvents::query()->get()->countBy('wiki_id')->sort()->reverse()->take(10)
= Illuminate\Support\Collection {#12823
    all: [
      457 => 282,
      216 => 260,
      209 => 196,
      326 => 173,
      516 => 171,
      647 => 115,
      501 => 113,
      448 => 107,
      694 => 106,
      385 => 99,
    ],
  }
```

m90:
WikiSiteStats (where the same problem was fixed in the same way in #785) looks better

```
> WikiSiteStats::query()->get()->countBy('wiki_id')->sort()->reverse()->take(10)
[!] Aliasing 'WikiSiteStats' to 'App\WikiSiteStats' for this Tinker session.
= Illuminate\Support\Collection {#17181
    all: [
      1250 => 1,
      1249 => 1,
      1248 => 1,
      1247 => 1,
      1246 => 1,
      1245 => 1,
      1244 => 1,
      1243 => 1,
      1241 => 1,
      1240 => 1,
    ],
  }
```

tests/Routes/Wiki/ConversionMetricTest.php
```diff
@@ -47,28 +47,6 @@ public function testDownloadCsv()
         $response->assertSee('two.wikibase.cloud');
     }
 
-    public function testDownloadJsonWithDuplicateLifecycleEvents() {
-        $current_date = CarbonImmutable::now();
-        $wiki = $this->createTestWiki('one.wikibase.cloud', 10, 9, 2);
-        $wiki->wikiLifecycleEvents()->create([
```
m90, line 53: This test now has to be removed as creating a duplicate entry will not work anymore due to the added database constraint.

deer-wmde:
looks solid to me for preventing duplicate entries, but I wonder if the code will now fail when attempting to do so or if it will "just work" the right way

deer-wmde (approved):

m90:
> but I wonder if the code will now fail when attempting to do so or if it will "just work" the right way

Tbh I don't know either (I would hope it works because of the lock), but if it fails it will point us to who is actually trying to create the duplicates now.

tarrow:
I just wanted to say I liked how this was done; when I wrote the ticket for this work I didn't think about doing the cleanup in the migration.

That is much nicer than what I had planned (and why I left it on the table) to manually cleanup and then quickly deploy and run the constraint migration.

PR #823, opened 2024-06-27
Add API endpoints for handling entity imports

m90:
Ticket: https://phabricator.wikimedia.org/T368021

This PR adds three new endpoints:

1. list entity imports for a wiki
2. create an entity import for a wiki
3. update an entity import for a wiki (expected to be called via backend routes by the k8s import job)

### DIFF
--- a/app/Http/Controllers/WikiEntityImportController.php
+++ b/app/Http/Controllers/WikiEntityImportController.php
@@ -6,6 +6,7 @@ use App\WikiEntityImportStatus;
 use Illuminate\Http\Request;
 use App\Wiki;
 use App\WikiEntityImport;
+use App\Jobs\WikiEntityImportJob;
 use Carbon\Carbon;
 
 class WikiEntityImportController extends Controller
@@ -41,10 +42,20 @@ class WikiEntityImportController extends Controller
                     ->setStatusCode(400);
             }
         }
-        $wiki->wikiEntityImports()->create([
+
+        $import = $wiki->wikiEntityImports()->create([
             'status' => WikiEntityImportStatus::Pending,
             'started_at' => Carbon::now(),
+            'payload' => $request->post(),
         ]);
+
+        dispatch(new WikiEntityImportJob(
+            wikiId: $wiki->id,
+            sourceWikiUrl: $request->input('source_wiki_url'),
+            importId: $import->id,
+            entityIds: $request->input('entity_ids'),
+        ));
+
         return response()->json(['data' => $import]);
     }
 
--- a/app/Http/Controllers/WikiEntityImportController.php
+++ b/app/Http/Controllers/WikiEntityImportController.php
@@ -6,7 +6,7 @@ use App\WikiEntityImportStatus;
 use Illuminate\Http\Request;
 use App\Wiki;
 use App\WikiEntityImport;
-use App\Jobs\WikiEntityImportJob;
+use App\Jobs\WikiEntityImportDummyJob;
 use Carbon\Carbon;
 
 class WikiEntityImportController extends Controller
@@ -49,7 +49,7 @@ class WikiEntityImportController extends Controller
             'payload' => $request->all(),
         ]);
 
-        dispatch(new WikiEntityImportJob(
+        dispatch(new WikiEntityImportDummyJob(
             wikiId: $wiki->id,
             sourceWikiUrl: $request->input('source_wiki_url'),
             importId: $import->id,
--- a/app/Http/Controllers/WikiEntityImportController.php
+++ b/app/Http/Controllers/WikiEntityImportController.php
@@ -1,0 +1,59 @@
+<?php
+
+namespace App\Http\Controllers;
+
+use App\WikiEntityImportStatus;
+use Illuminate\Http\Request;
+use App\Wiki;
+use App\WikiEntityImport;
+use Carbon\Carbon;
+
+class WikiEntityImportController extends Controller
+{
+    public function get(Request $request): \Illuminate\Http\JsonResponse
+    {
+        $wiki = Wiki::find($request->input('wiki_id'));
+        if (!$wiki) {
+            return response()->json(['error' => 'Not Found'])->setStatusCode(404);
+        }
+        $imports = $wiki->wikiEntityImports()->get();
+        return response()->json(['data' => $imports]);
+    }
+
+    public function create(Request $request): \Illuminate\Http\JsonResponse
+    {
+        $wiki = Wiki::find($request->input('wiki_id'));
+        if (!$wiki) {
+            return response()->json(['error' => 'Not Found'])->setStatusCode(404);
+        }
+        $imports = $wiki->wikiEntityImports()->get();
+        foreach ($imports as $import) {
+            if ($import->status === WikiEntityImportStatus::Success) {
+                return response()
+                    ->json(['error' => 'Wiki "'.$wiki->domain.'" already has performed a successful entity import'])
+                    ->setStatusCode(400);
+            }
+            if ($import->status === WikiEntityImportStatus::Pending) {
+                return response()
+                    ->json(['error' => 'Wiki "'.$wiki->domain.'" currently has a pending entity import'])
+                    ->setStatusCode(400);
+            }
+        }
+        $import = WikiEntityImport::create([
+            'wiki_id' => $wiki->id,
+            'status' => WikiEntityImportStatus::Pending,
+            'started_at' => Carbon::now(),
+        ]);
+        return response()->json(['data' => $import]);
+    }
+
+    public function update(Request $request): \Illuminate\Http\JsonResponse
+    {
+        $import = WikiEntityImport::find($request->input('wiki_entity_import_id'));
+        $import->update([
+            'status' => $request->input('status'),
+            'finished_at' => Carbon::now(),
+        ]);
+        return response()->json(['data' => $import]);
+    }
+}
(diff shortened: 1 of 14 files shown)


database/migrations/2024_06_24_074114_create_wiki_entity_imports_table.php
```diff
@@ -0,0 +1,35 @@
...
+            $table->id();
+            $table->timestamps();
+
+            $table->unsignedInteger('wiki_id');
+            $table->foreign('wiki_id')->references('id')->on('wikis');
+
+            $table->enum('status', ['pending', 'failed', 'success']);
+            $table->timestamp('started_at')->nullable();
+            $table->timestamp('finished_at')->nullable();
+            $table->json('payload')->nullable();
```
m90, line 24: Technically, storing this information is not necessary, but I thought it might be very helpful when debugging failed runs while rolling out this feature. Once things work, we might think about dropping it again.

app/Jobs/WikiEntityImportJob.php
```diff
@@ -0,0 +1,34 @@
...
+        public string $importId,
+    )
+    {}
+
+    /**
+     * Execute the job.
+     */
+    public function handle(): void
+    {
+        Log::info("Handling job for ".$this->wikiId.", ".$this->sourceWikiUrl.", ".implode(",", $this->entityIds)." ,".$this->importId);
```
m90, line 32: Obviously, this job currently is just a dud. Its proper implementation is covered by https://phabricator.wikimedia.org/T368024

app/Http/Middleware/LimitWikiAccess.php
```diff
@@ -0,0 +1,30 @@
...
+{
+    /**
+     * Handle an incoming request.
+     *
+     * @param  \Closure(\Illuminate\Http\Request): (\Symfony\Component\HttpFoundation\Response)  $next
+     */
+    public function handle(Request $request, Closure $next): Response
+    {
+        $userHasAccess = WikiManager::where([
+            'user_id' => $request->user()?->id,
```
deer-wmde, line 20: so happy to learn about this operator, perfect use for it as well. :clap:

app/Http/Middleware/LimitWikiAccess.php
```diff
@@ -0,0 +1,30 @@
...
+            'wiki_id' => $request->input('wiki'),
```
deer-wmde, line 21: do we need validation/sanitation here for the two user inputs?

database/migrations/2024_06_24_074114_create_wiki_entity_imports_table.php
deer-wmde: it might even be useful to display this information to the user, in case they forgot or are confused if they did the importing and on which wiki, etc.

routes/api.php
```diff
@@ -46,6 +43,11 @@
             $router->post('setting/{setting}/update', ['uses' => 'WikiSettingController@update']);
             // TODO should wiki managers really be here?
             $router->post('managers/list', ['uses' => 'WikiManagersController@getManagersOfWiki']);
+            $router->get('entityImport', ['middleware' => 'limit_wiki_access', 'uses' => 'WikiEntityImportController@get']);
+            $router->post('entityImport', ['middleware' => 'limit_wiki_access', 'uses' => 'WikiEntityImportController@create']);
```
deer-wmde, line 47: is the third endpoint (update) missing here?

routes/backend.php
```diff
@@ -15,8 +15,8 @@ function () {
 );
 
 $router->group(['prefix' => 'wiki'], function () use ($router) {
-    // GET
     $router->get('getWikiForDomain', ['uses' => 'WikiController@getWikiForDomain']);
+    $router->patch('updateEntityImport', ['uses' => '\App\Http\Controllers\WikiEntityImportController@update']);
```
deer-wmde, line 19: is http `PATCH` used here because of semantics? not against it, just wondering because I'm not sure I encountered it in this repo before

app/Http/Controllers/WikiEntityImportController.php
```diff
@@ -0,0 +1,87 @@
...
+            importId: $import->id,
+            entityIds: $request->input('entity_ids'),
+        ));
+
+        return response()->json(['data' => $import]);
+    }
+
+    public function update(Request $request): \Illuminate\Http\JsonResponse
+    {
+        // This route is not supposed have ACL middlewares in front as it is expected
```
deer-wmde, line 64: great comment :star:

app/WikiEntityImport.php
```diff
@@ -0,0 +1,31 @@
...
+    protected $fillable = self::FIELDS;
+
+    protected $visible = self::FIELDS;
+
+    protected $casts = [
+        'status' => WikiEntityImportStatus::class,
+    ];
+}
+
+enum WikiEntityImportStatus: string
```
deer-wmde, line 26: love the use of enum here

app/Http/Controllers/WikiEntityImportController.php
```diff
@@ -0,0 +1,87 @@
...
+        $wiki = Wiki::find($request->input('wiki'));
+        if (!$wiki) {
+            abort(404, 'No such wiki');
+        }
+
+        $imports = $wiki->wikiEntityImports()->get();
+        foreach ($imports as $import) {
+            if ($import->status === WikiEntityImportStatus::Success) {
+                return response()
+                    ->json(['error' => 'Wiki "'.$wiki->domain.'" already has performed a successful entity import'])
```
deer-wmde, line 36: I'm not sure I understand this logic (or rather the intended function) right: is the idea that per wiki only one import is gonna be possible?

app/Http/Controllers/WikiEntityImportController.php
```diff
@@ -0,0 +1,87 @@
...
+                return response()
+                    ->json(['error' => 'Wiki "'.$wiki->domain.'" currently has a pending entity import'])
+                    ->setStatusCode(400);
+            }
+        }
+
+        $import = $wiki->wikiEntityImports()->create([
+            'status' => WikiEntityImportStatus::Pending,
+            'started_at' => Carbon::now(),
+            'payload' => $request->all(),
```
deer-wmde, line 49: I'm curious about the structure of `payload` here,  do we leave it intentionally flexible?

app/Http/Controllers/WikiEntityImportController.php
```diff
@@ -0,0 +1,87 @@
...
+            wikiId: $wiki->id,
+            sourceWikiUrl: $request->input('source_wiki_url'),
+            importId: $import->id,
+            entityIds: $request->input('entity_ids'),
+        ));
+
+        return response()->json(['data' => $import]);
+    }
+
+    public function update(Request $request): \Illuminate\Http\JsonResponse
```
deer-wmde, line 62: is the idea of this method that we have a mechanism to stop an import?

app/Http/Middleware/LimitWikiAccess.php
m90: I would really hope Laravel does not let me pass user provided SQL to a QueryBuilder, but just for the sake of better error messages validation probably makes sense. Will add this.

routes/backend.php
m90: Yeah, mostly, you can read using GET, create using POST and update using PATCH.

app/Http/Controllers/WikiEntityImportController.php
m90: Yes. This is what the ticket says

> the button to copy the entities can only be triggered once per wiki. after having copied them over, the button is disabled and a message appears informing the user that copying can only happen once

I'm not sure why that is though.

app/Http/Controllers/WikiEntityImportController.php
m90: Right now it's mostly a debugging helper, so I thought it'd be good to make it as flexible as possible.

app/Http/Controllers/WikiEntityImportController.php
m90: The idea is that the job can signal back the outcome of its work. I.e. the user POSTs an update via UI, a job is created, and then the job can use the backend route to tell API whether the run succeeded or failed.

app/Http/Middleware/LimitWikiAccess.php
m90: Added this :heavy_check_mark:

deer-wmde: > I would really hope Laravel does not let me pass user provided SQL to a QueryBuilder

I'd be quite surprised if Laravel would circumvent that actually (not that it wouldnt be a great idea, I just dont have high hopes here :D)

app/Http/Controllers/WikiEntityImportController.php
deer-wmde: Aha interesting! Okay then at least I didn't misunderstand it

app/Http/Controllers/WikiEntityImportController.php
deer-wmde: ah right, got it, thanks

app/Http/Middleware/LimitWikiAccess.php
m90: I think unless you use `raw` methods, this is prevented by Eloquent https://www.stackhawk.com/blog/sql-injection-prevention-laravel/

deer-wmde (approved):

m90:
Changed the base branch for this to a long running feature branch kind of thing that collects all of the follow ups as well.

PR #825, opened 2024-06-28
Save reason for deleting wikis

rosalieper:


app/Http/Controllers/WikiController.php
```diff
@@ -167,18 +167,23 @@ public function delete(Request $request): \Illuminate\Http\JsonResponse
...
 
         // Check that the requesting user manages the wiki
         if (WikiManager::where('user_id', $userId)->where('wiki_id', $wikiId)->count() !== 1) {
             // The deletion was requested by a user that does not manage the wiki
             return response()->json('Unauthorized', 401);
         }
+        $wiki = Wiki::find($wikiId);
+
+        //Save the wiki deletion reason
+        $wiki->update(['wiki_deletion_reason' => $wikiDeletionReason]);
```
m90, line 180: Question: should this be skipped in case no deletion reason is given in the request?

rosalieper: yes, I made an update to that. thank you

PR #830, opened 2024-07-03
Run `transferbot` in a Kubernetes Job

m90:
Ticket https://phabricator.wikimedia.org/T368024

Depends on https://github.com/wbstack/mediawiki/pull/447, https://github.com/wbstack/transferbot/pull/3


app/Http/Controllers/WikiEntityImportController.php
```diff
@@ -30,7 +30,7 @@ public function create(Request $request): \Illuminate\Http\JsonResponse
     {
         $validatedInput = $request->validate([
             'wiki' => ['required', 'integer'],
-            'source_wiki_url' => ['required', 'string'],
+            'source_wiki_url' => ['required', 'url'],
```
m90, line 33: This is a drive by fix that prevents a potential code injection vector. The value is passed to the job's command, so it's important this does not contain potential code snippets we don't want to run. A well formed URL should be safe from what I understand.

app/Jobs/WikiEntityImportJob.php
```diff
@@ -0,0 +1,240 @@
...
+     */
+    public function handle(Client $kubernetesClient): void
+    {
+        $import = null;
+        try {
+            $wiki = Wiki::findOrFail($this->wikiId);
+            $import = WikiEntityImport::findOrFail($this->importId);
+            $creds = $this->acquireCredentials($wiki->domain);
+
+            $this->targetWikiUrl = str_contains($wiki->domain, "localhost")
```
m90, line 48: This is ugly. I'd be happy to learn about a cleaner solution for this.

app/Jobs/WikiEntityImportJob.php
```diff
@@ -0,0 +1,240 @@
...
+
+    private static function acquireCredentials(string $wikiDomain): OAuthCredentials
+    {
+        $response = Http::withHeaders(['host' => $wikiDomain])->asForm()->post(
+            getenv('PLATFORM_MW_BACKEND_HOST').'/w/api.php?action=wbstackPlatformOauthGet&format=json',
+            [
+                'consumerName' => 'WikiEntityImportJob',
+                'ownerOnly' => '1',
+                'consumerVersion' => '1',
+                'grants' => 'basic|highvolume|import|editpage|editprotected|createeditmovepage|uploadfile|uploadeditmovefile|rollback|delete|mergehistory',
```
m90, line 85: I'm pretty sure these could be reduced even further, but I am totally lost when debugging and was happy I found a working combo that is not "every permission ever".

app/Jobs/WikiEntityImportJob.php
```diff
@@ -0,0 +1,240 @@
...
+                'generateName' => 'run-transferbot-',
+                'namespace' => $this->kubernetesNamespace,
+                'labels' => [
+                    'app.kubernetes.io/instance' => $this->wiki->domain,
+                    'app.kubernetes.io/name' => 'run-transferbot',
+                ]
+            ],
+            'spec' => [
+                'ttlSecondsAfterFinished' => 0,
+                'backoffLimit' => 0,
```
m90, line 150: This means we only ever retry once and leave doing this up to the user that triggered a failed job.

app/Jobs/WikiEntityImportJob.php
```diff
@@ -0,0 +1,240 @@
...
+                'template' => [
+                    'metadata' => [
+                        'name' => 'run-transferbot'
+                    ],
+                    'spec' => [
+                        'containers' => [
+                            0 => [
+                                'hostNetwork' => true,
+                                'name' => 'run-qs-updater',
+                                'image' => 'ghcr.io/wbstack/transferbot:main',
```
m90, line 160: TODO: this should still refer to a tagged version.

app/Jobs/WikiEntityImportJob.php
```diff
@@ -0,0 +1,240 @@
...
+                                'resources' => [
+                                    'requests' => [
+                                        'cpu' => '0.25',
+                                        'memory' => '250Mi',
+                                    ],
+                                    'limits' => [
+                                        'cpu' => '0.5',
+                                        'memory' => '500Mi',
+                                    ],
+                                ],
```
m90, line 187: This is not even an educated guess. I would like to see this run in staging against whatever we decide to import before coming up with real world numbers.

app/Http/Controllers/WikiEntityImportController.php
tarrow: good catch!

app/Jobs/WikiEntityImportJob.php
tarrow: only suggestion I have is we revisit the behaviour of the TransferBot image *again* and do something like the sidecar proxy thing and always talk to that over http.

I agree this is a nasty leaking of knowing the details about the network environment of the job container into the api.

We could at least tighten it by parsing the url and checking it's the TLD of the FQDN. That would save us having some mystery errors in the future for `localhostingofdinnerparties.wikibase.cloud` which this would inadvertently catch

app/Jobs/WikiEntityImportJob.php
tarrow: I'm surprised most of these are needed. I'm trying to think if there is a quick way for us to test these

app/Jobs/WikiEntityImportJob.php
tarrow: I think generally keeping the request low and the limit high here will be fine so it can burst to whatever is available. We con't really need this thing to run "realtime" like stuff that's replying to webrequests but some real world numbers will probably make us happy

app/Jobs/WikiEntityImportJob.php
```diff
@@ -0,0 +1,240 @@
...
+                'backoffLimit' => 0,
+                'template' => [
+                    'metadata' => [
+                        'name' => 'run-transferbot'
+                    ],
+                    'spec' => [
+                        'containers' => [
+                            0 => [
+                                'hostNetwork' => true,
+                                'name' => 'run-qs-updater',
```
tarrow, line 159: I'd probably change this from the copy paste. 

Indeed inf we are looking for a name I'd be keen on `EntityImport` and moving away from transferbot (some random chat thread we had one month ago: https://mattermost.wikimedia.de/swe/pl/4kao4tr49inobd4i45iuz8cqqo)

app/Jobs/WikiEntityImportJob.php
```diff
@@ -0,0 +1,240 @@
...
+                                'image' => 'ghcr.io/wbstack/transferbot:1.0.0',
```
tarrow, line 160: I'd be tempted to get this version number in an variable since I can imagine us poking it ofter

app/Jobs/WikiEntityImportJob.php
m90: Yeah, I think parsing the domain and then checking its TLD is a good compromise.

app/Jobs/WikiEntityImportJob.php
m90: Apart from the typo, my reasoning for using transferbot wording here was that this is the name of the image we're running. But maybe ditching this in favor of entity-import is better, yes.

app/WikiEntityImportStatus.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace App;
+
+enum WikiEntityImportStatus: string
```
tarrow, line 5: praise: look at use using the fancy language features

app/Jobs/WikiEntityImportJob.php
```diff
@@ -0,0 +1,250 @@
...
+
+    private static function domainToOrigin(string $domain): string
+    {
+        $tld = last(explode('.', $domain));
+        return $tld === 'localhost'
+            ? "http://".$domain
+            : "https://".$domain;
+    }
+
+    private static function acquireCredentials(string $wikiDomain): OAuthCredentials
```
tarrow, line 83: comment: after the testing thoughts yesterday I was left wondering if I should be recommending moving this logic into the transferbot image but I think that it probably still makes most sense here. This is slightly different to QuickStatements et al. but given that all the import stuff is now split between here and the `transferbot` image I think it's a fair argument to keep the content of the transferbot image fairly minimal

m90: It would also mean the transferbot image would be super tightly bound to wbstack and can't even be used in any other setup. It would be a thing we could do, but if we do it, we might also de-generalize other parts in there (i.e. selecting wikis by Host header and such).

tarrow (approved):

PR #839, opened 2024-07-16
feat(db): add more seeded wikibases

AndrewKostka:


database/seeds/WikisSeeder.php
```diff
@@ -1,44 +1,89 @@
...
 
 class WikisSeeder extends Seeder
 {
-    public function run()
+    /**
+     * @param \App\User $user
+     * @param string $name
+     * @param array $stats
+     */
+    public function createWiki($user, $name, $stats)
```
tarrow, line 23: Would this be better as private? Are you expecting to call it anywhere other than `$this->run()`?

database/seeds/WikisSeeder.php
```diff
@@ -1,44 +1,89 @@
...
+    public function run()
+    {
+        $email = 'seeder@email.example';
+        $user = User::create([
+            'email' => $email,
+            'password' => Hash::make($email),
+            'verified' => true
         ]);
+
+        for ($id = 0; $id < 50; $id++) {
```
tarrow, line 85: This could end up being misleading since currently we have a production limit of 6 wikis per user so adding 50 using internal methods might make things appear unusual. Would it make sense to also create 50 users?

database/seeds/WikisSeeder.php
```diff
@@ -1,44 +1,89 @@
...
         WikiDomain::create([
-            'domain' => 'seededsite.nodomain.dev',
-            'wiki_id' => $wiki->id,
+            'domain' => $domain,
+            'wiki_id' => $wiki->id
+        ]);
+
+        $index = substr(bin2hex(random_bytes(48)), 0, 10);
+
+        WikiDb::create([
```
tarrow, line 39: This looks much nicer and more consistent than then `DB::table` calls

database/seeds/WikisSeeder.php
```diff
@@ -1,44 +1,89 @@
...
         WikiSetting::create([
-            'wiki_id' => $wiki->id,
             'name' => 'wgSecretKey',
             'value' => Str::random(64),
+            'wiki_id' => $wiki->id
         ]);
 
         WikiSetting::create([
-            'wiki_id' => $wiki->id,
             'name' => WikiSetting::wwExtEnableElasticSearch,
```
dati18, line 61: I don't know if it's necessary to enable ES? Is it needed somewhere?

database/seeds/WikisSeeder.php
```diff
@@ -1,44 +1,89 @@
...
+    {
+        $email = 'seeder@email.example';
+        $user = User::create([
+            'email' => $email,
+            'password' => Hash::make($email),
+            'verified' => true
         ]);
+
+        for ($id = 0; $id < 50; $id++) {
+            $this->createWiki($user, 'seededsite-' . $id, [ 'pages' => $id ]);
```
dati18, line 86: As Tom asked before, I don't know if it's necessary to createUser  also

database/seeds/WikisSeeder.php
```diff
@@ -1,44 +1,89 @@
...
+        WikiDb::create([
+            'name' => 'mwdb_' . $index,
+            'user' => 'mwu_' . $index,
+            'password' => $index,
+            'version' => Config::get('wbstack.wiki_db_use_version'),
+            'prefix' => 'mwt_' . $index,
+            'wiki_id' => $wiki->id
+        ]);
+
+        QueryserviceNamespace::create([
```
dati18, line 48: My gut feeling said that we only need to create the simplest of wikis (meaning name, user, version, id... etc)
I don't know if we really need other things? We only need it for testing and other stuff is just redundant. Unless you want to use the in other cases

dati18: Frederik used to bashed me for adding things that we "will use in the future but not now" and the future sometimes come sooner than we expected.
So I will agree to keep them here. It doesn't hurt but I will keep it in mind

dati18 (approved):

PR #840, opened 2024-07-17
Test the csv output of DeletedWikiMetricsController

tarrow:
This commit attempts to directly test the output of the DeletedWikiMetricsController by creating a user that has the ability to download the CSV and then parsing using the php csv builtin[1].

We then assert some values in this

[1] https://www.php.net/manual/en/function.str-getcsv.php

### DIFF
--- a/tests/Routes/Wiki/DeletedWikiMetricsControllerTest.php
+++ b/tests/Routes/Wiki/DeletedWikiMetricsControllerTest.php
@@ -72,12 +72,17 @@ class DeletedWikiMetricsControllerTest extends TestCase
             $this->createAndDeleteTestWiki('two.wikibase.cloud', $user2->id, 'Some Reason', 0, 3),
             $this->createAndDeleteTestWiki('sameuser.wikibase.cloud', $user1->id, 'Some Reason', 0, 3),
         ];
-        $controller = new DeletedWikiMetricsController();
-        $output = $controller->createOutput($deletedWikis);
-        $this->assertSame('one.wikibase.cloud', $output[0]['domain_name_for_wiki']);
-        $this->assertSame('two.wikibase.cloud', $output[1]['domain_name_for_wiki']);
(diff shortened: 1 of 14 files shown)


tarrow:
As you can see this fails because there is a deleted wiki that wasn't cleaned up by a previous test.

I started investigating where this could be and found it likely that multiple tests are polluting the test environment despite attempts like `RefreshDatabase` and `DatabaseTransactions`

I did this by the ugly hack of creating a global in our base testcase set to the name of the test and including this in the WikiFactory domain name.

tests/Routes/Wiki/DeletedWikiMetricsControllerTest.php
```diff
@@ -72,12 +72,17 @@ public function testOutputHasCorrectContent()
...
-        $this->assertSame('two.wikibase.cloud', $output[1]['domain_name_for_wiki']);
-        $this->assertSame('Some Reason',$output[1]['wiki_deletion_reason']);
-        $this->assertSame(2, $output[0]['number_of_wikibases_owned_by_owners_of_this_wiki']);
+        $user = $this->createUserWithPrivileges(1);
+        $this->actingAs($user, 'api')->get('auth/login');
+
+        $response = $this->get($this->route);
+
+        $rawCsv = $response->getContent();
+        $output = array_map('str_getcsv', explode("\n", $rawCsv));
```
rosalieper, line 81: This is way better. thank you @tarrow

rosalieper:
I don't know how I feel about skipping this many tests. Do we want to revisit them later or we just don't want to run them anymore.

tarrow:
yeah, totally fair point. I'll take a short look at trying to make these less evil without investing a huge amount of time. If we if we leave them skipped we certainly want to revisit them later but I think you are right to be nervous; a skipped test now will become broken a janky before you know it and the work to resurrect it could be hard

rosalieper (approved):

PR #847, opened 2024-07-23
feat: emit metrics about entity imports

m90:
Ticket https://phabricator.wikimedia.org/T370799

### DIFF
--- a/app/Http/Controllers/WikiEntityImportController.php
+++ b/app/Http/Controllers/WikiEntityImportController.php
@@ -9,9 +9,26 @@ use App\Wiki;
 use App\WikiEntityImport;
 use App\Jobs\WikiEntityImportJob;
 use Carbon\Carbon;
+use LKDevelopment\HorizonPrometheusExporter\Repository\ExporterRepository;
+use Prometheus\Counter;
 
 class WikiEntityImportController extends Controller
 {
+    private Counter $successfulCounter;
+    private Counter $failedCounter;
+
+    public function __construct()
+    {
+        ExporterRepository::load();
+        $this->successfulCounter = ExporterRepository::getRegistry()->getCounter(
+            config('horizon-exporter.namespace'),
+            'wiki_entity_imports_successful',
+        );
+        $this->failedCounter = ExporterRepository::getRegistry()->getCounter(
+            config('horizon-exporter.namespace'),
+            'wiki_entity_imports_failed',
+        );
+    }
     public function get(Request $request): \Illuminate\Http\JsonResponse
     {
         $validatedInput = $request->validate([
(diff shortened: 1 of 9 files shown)


app/Metrics/WikiEntityImports.php
```diff
@@ -0,0 +1,49 @@
...
+    protected $failed;
+
+    public function metrics(CollectorRegistry $collectorRegistry)
+    {
+        $this->pending = $collectorRegistry->getOrRegisterGauge(
+            config('horizon-exporter.namespace'),
+            'wiki_entity_imports_pending',
+            'The number of pending Entity imports currently being processed.',
+        );
+        $this->successful = $collectorRegistry->getOrRegisterGauge(
```
m90, line 24: It's a bit unfortunate that both this as well as `failed` are gauges when instead they should be counters. However, the library in use does not expose an option for calling SET on a counter, instead it only knows about INCREASE, which would require knowledge about the previous state in this scenario.

tarrow: Could we inject a collector registry into the controller and call `increase` there?

https://github.com/wbstack/api/pull/830/files#diff-29372c16f0d76454ce93dfa22a7e0833e4ae2a7a404e3adee053364af885a060R79

I'm not sure if this is excessively abusing the "horizon" nature of the exporter but this would be more "natural" than a gauge perhaps. Although, I think given that currently we don't have a plan to garbage collect old Statuses it's not that bad to treat this as a gauge

m90: Should be worth a try.

m90: This is now implemented.

app/Http/Controllers/WikiEntityImportController.php
```diff
@@ -9,9 +9,27 @@
...
+use Prometheus\Counter;
 
 class WikiEntityImportController extends Controller
 {
+    private Counter $successfulCounter;
+    private Counter $failedCounter;
+
+    public function __construct(CollectorRegistry $registry)
+    {
+        $this->successfulCounter = $registry->getOrRegisterCounter(
```
m90, line 22: There's a tricky decision here: if we define these counters only once, it means they will be missing from the metrics if they have never been increased yet (the controller that answers the scraper will not know about it yet as it's never been persisted).

Alternatively we could:
- duplicate this definition in the `Metrics/WikiEntityImport` class
- move all metrics definitions to the service provider and have stub `metrics` methods on the metrics classes themselves

config/database.php
```diff
@@ -157,6 +157,12 @@
             'database' => env('REDIS_CACHE_DB', 1),
         ],
 
+        'metrics' => [
+            'host' => env('REDIS_HOST', '127.0.0.1'),
+            'password' => env('REDIS_PASSWORD', null),
+            'port' => env('REDIS_PORT', 6379),
+            'database' => env('REDIS_METRICS_DB', 2),
```
m90, line 164: This is not supported by the Prometheus library that just uses a hardcoded `PROMETHEUS_` prefix on keys. I shall remove the setting again.

app/Http/Controllers/WikiEntityImportController.php
tarrow: If I understand right this isn't really a big problem? It just means that instead of `0` we will have `null` until the first import?

tarrow (approved):

tarrow (approved):

PR #853, opened 2024-07-31
chore: allow for revisioned entity ids in validation

m90:
Connects https://github.com/wbstack/transferbot/pull/6

When we want to pass revisioned ids to transferbot, current validation rules are too strict and need to allow for a potential `@123` suffix.

### DIFF
--- a/app/Http/Controllers/WikiEntityImportController.php
+++ b/app/Http/Controllers/WikiEntityImportController.php
@@ -34,7 +34,7 @@ class WikiEntityImportController extends Controller
             'entity_ids' => ['required', 'string', function (string $attr, mixed $value, \Closure $fail) {
                 $chunks = explode(',', $value);
                 foreach ($chunks as $chunk) {
-                    if (!preg_match("/^[A-Z]\d+$/", $chunk)) {
+                    if (!preg_match("/^[A-Z]\d+(@\d+)?$/", $chunk)) {
                         $fail("Received unexpected input '{$chunk}' cannot continue.");
                     }
                 }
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/WikiEntityImportController.php
```diff
@@ -34,7 +34,7 @@ public function create(Request $request): \Illuminate\Http\JsonResponse
             'entity_ids' => ['required', 'string', function (string $attr, mixed $value, \Closure $fail) {
                 $chunks = explode(',', $value);
                 foreach ($chunks as $chunk) {
-                    if (!preg_match("/^[A-Z]\d+$/", $chunk)) {
+                    if (!preg_match("/^[A-Z]\d+(@\d+)?$/", $chunk)) {
```
m90, line 37: Just double checking: is my understanding correct that revision ids will always be integers and nothing else?

tarrow: correct :)

tests/Routes/Wiki/EntityImportTest.php
```diff
@@ -154,7 +154,7 @@ public function testCreateValidation()
             ->assertStatus(422);
 
         $this->assertEquals(0, WikiEntityImport::count());
-        Bus::assertDispatchedTimes(WikiEntityImportDummyJob::class, 0);
+        Bus::assertDispatchedTimes(WikiEntityImportJob::class, 0);
```
tarrow, line 157: good catch; naturally neither the Dummy nor the Real job ran at all

tarrow (approved):

PR #854, opened 2024-07-31
Normalised first and lasted edited time to use revisions

tarrow:
This commit calculates the last edited time using a similar methodology to the existing on to calculated first edited time. It uses the `allrevisions` list.

This method will exclude all revisions created by the PlatformReservedUser. This means that it will erroneously ignore revisions from importing entities.

Bug: T364991

### DIFF
--- a/database/migrations/2024_08_01_193038_remove_and_rebuild_wiki_lifecycle_events.php
+++ b/database/migrations/2024_08_01_193038_remove_and_rebuild_wiki_lifecycle_events.php
@@ -1,0 +1,25 @@
+<?php
+
+use App\Jobs\UpdateWikiSiteStatsJob;
+use App\WikiLifecycleEvents;
+use Illuminate\Database\Migrations\Migration;
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        WikiLifecycleEvents::all()->map->delete();
+        (new UpdateWikiSiteStatsJob())->handle();
+    }
+
+    /**
+     * Reverse the migrations.
+     */
+    public function down(): void
+    {
+        //
+    }
+};
--- a/database/migrations/2024_08_01_193038_remove_and_rebuild_wiki_lifecycle_events.php
+++ b/database/migrations/2024_08_01_193038_remove_and_rebuild_wiki_lifecycle_events.php
@@ -12,7 +12,7 @@ return new class extends Migration
     public function up(): void
     {
         WikiLifecycleEvents::query()->delete();
-        (new UpdateWikiSiteStatsJob())->handle();
+        UpdateWikiSiteStatsJob::dispatch();
     }
 
     /**
--- a/database/migrations/2024_08_01_193038_remove_and_rebuild_wiki_lifecycle_events.php
+++ b/database/migrations/2024_08_01_193038_remove_and_rebuild_wiki_lifecycle_events.php
@@ -11,7 +11,7 @@ return new class extends Migration
      */
     public function up(): void
     {
-        WikiLifecycleEvents::all()->map->delete();
+        WikiLifecycleEvents::query()->delete();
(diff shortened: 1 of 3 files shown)


database/migrations/2024_08_01_193038_remove_and_rebuild_wiki_lifecycle_events.php
```diff
@@ -0,0 +1,25 @@
...
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        WikiLifecycleEvents::all()->map->delete();
+        (new UpdateWikiSiteStatsJob())->handle();
```
m90, line 15: Does this run synchronously and waits until all jobs for all wikis are finished or will it just dispatch a set of jobs into the queue and then move on? Asking because I would think calling `handle` on a job is a wee bit of an anti pattern, but maybe we just need it.

database/migrations/2024_08_01_193038_remove_and_rebuild_wiki_lifecycle_events.php
```diff
@@ -0,0 +1,25 @@
...
+use Illuminate\Database\Migrations\Migration;
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        WikiLifecycleEvents::all()->map->delete();
```
m90, line 14: Is `map` expected to be here?

database/migrations/2024_08_01_193038_remove_and_rebuild_wiki_lifecycle_events.php
tarrow: > Does this run synchronously and waits until all jobs for all wikis are finished

It does this exactly; runs them synchronously and if they fail the migration will fail. You think it would be better to just dispatch it to the queue?

database/migrations/2024_08_01_193038_remove_and_rebuild_wiki_lifecycle_events.php
tarrow: It was deliberate but it's perhaps a somewhat confusing shorthand.

You can't simply call `delete()` on a collection but need to all it on each item in the collection. You can read about it here https://laravel.com/docs/10.x/collections#higher-order-messages

Do you think this is too obscure to be understandable and the more simple
```
->map(function (int $item, int $key) {
    $item->delete();
});
```

would be nicer?

m90: In the tests I usually use `Something::query()->delete()` which seems to work as intended. Any reason against using this? The map as is is indeed slightly confusing?

database/migrations/2024_08_01_193038_remove_and_rebuild_wiki_lifecycle_events.php
m90: I would think that the migration runs in a transaction, and this means we have a transaction open for a very long time. But maybe that's ok.

database/migrations/2024_08_01_193038_remove_and_rebuild_wiki_lifecycle_events.php
tarrow: yeah, that's definitely way better; just hadn't thought of it. It's probably also "more efficient" since it doesn't trigger loading all the Events into memory

database/migrations/2024_08_01_193038_remove_and_rebuild_wiki_lifecycle_events.php
tarrow: yeah, I would guess it be open for best part of a minute+. I've made the job dispatchable and instead dispatched it.

To paste from my local setup this of 50 Wikis from ~10s to 32ms to run the migration

m90 (approved):

PR #856, opened 2024-08-02
Fix typo on request input key

rosalieper:


deer-wmde (approved):

tarrow:
I clicked re-run after the fix in #857 was merged. Looks like this will still need a little something probably in DeleteWikiTest.php to fix up the probably duplicated typo there :)

m90 (approved):

PR #857, opened 2024-08-06
fix: compose is a docker plugin now

m90:
Starting today, tests started failing with

```
/home/runner/work/_temp/b56fec0a-4ac0-48be-9695-939a2d709bb9.sh: line 1: docker-compose: command not found
```

which apparently means compose v1 was now dropped from the runners.


tarrow:
I saw this happen a couple of times before; I suspect some runners still have it and others don't. I guess we're still gracefully rolling out to the newer runners. See https://github.com/actions/runner-images/issues/9692 and the rollout of https://github.com/actions/runner-images/issues/9848

tarrow (approved):

PR #871, opened 2024-10-28
workflow: fix ArgoCD update in build step

AndrewKostka:


tarrow:
This is necessary because we want to check if wbaas-deploy should be updated: i.e. https://github.com/wbstack/api/pull/871/files#diff-c59f4f9a90c4b34cef2c040c881f906daad13fde73566f0e34954f46b0f6efa3R57

tarrow (approved):

PR #875, opened 2024-11-18
Fix: Change email signature

dati18:
Bug: T374791


tarrow (approved):
lgtm, I also grepped for other `https://wikiba.se/contact/` instances and didn't see any :)

PR #883, opened 2025-02-28
Fix bug on wiki metrics recording

rosalieper:


tarrow (commented):
should we add a test for this case?

tests/Metrics/WikiMetricsTest.php
```diff
@@ -49,5 +49,32 @@ public function testDoesNotAddDuplicateRecordsWithOnlyDateChange()
...
+            'pages' => 0,
+            'is_deleted' => 0
+        ]);
+        //delete the wiki
+        $wiki->delete();
+        $wiki->save();
+
+        (new WikiMetrics())->saveMetrics($wiki);
+
+        //Assert No new record was created for today
```
tarrow, line 72: doesn't this assert that a new record *was* created for today?

tarrow: looks great other than this comment; please go ahead with it :)

deer-wmde: > doesn't this assert that a new record was created for today?

I think that's true. As the expected behaviour [seems to be](https://phabricator.wikimedia.org/T383421#10586021) that in case a wiki gets deleted one final row should get added where `is_deleted = 1`, I adjusted this test to first insert a row where exactly that happens "yesterday" but then no new row gets added "today"

PR #905, opened 2025-03-20
Create WikiProfile objects when a Wiki is made

tarrow:
This creates a WikiProfile object if a profile is provided
at wiki creation time.

It also add some validation for the form of this profile
over the network.

It should be noted that the _other strings are stored
unsantised user input. We need to take this into account
if we do anything with them

Bug: T389063

### DIFF
--- a/app/Helper/ProfileValidator.php
+++ b/app/Helper/ProfileValidator.php
@@ -1,0 +1,21 @@
+<?php
+namespace App\Helper;
+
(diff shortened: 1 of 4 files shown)


rosalieper (approved):

addshore:
Neat, will this be exposed publically at all?

tarrow:
> Neat, will this be exposed publically at all?

I would imagine so

PR #907, opened 2025-03-24
Add metrics: daily, weekly, monthly and quarterly number of actions

rosalieper:


rosalieper:
there is quite some repetition here. I will welcome any suggestions to make it better. I tried switch case and it only made it more complex and confusing.

rosalieper:
closing this pull request because it does not solve the task T383427. I just realized I misunderstood the whole task. :(

PR #909, opened 2025-03-25
bug: fix WikiProfile audience not nullable

tarrow:
WikiProfile audiences should be nullable but aren't I was unable to make them nullable using the laravel ORM

This migration does it using raw SQL following this issue [0] in the laravel bug tracker

[0] https://github.com/laravel/framework/issues/35096#issuecomment-1512754452

### DIFF
--- a/database/migrations/2025_03_25_085142_wiki_profiles_audience_nullable.php
+++ b/database/migrations/2025_03_25_085142_wiki_profiles_audience_nullable.php
@@ -1,0 +1,33 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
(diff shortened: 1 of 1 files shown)


database/migrations/2025_03_25_085142_wiki_profiles_audience_nullable.php
```diff
@@ -0,0 +1,33 @@
...
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+            DB::statement('
+                ALTER TABLE `wiki_profiles` CHANGE COLUMN `audience` `audience` 
+                ENUM("narrow", "wide", "other")
```
deer-wmde, line 16:
```suggestion
                ENUM("narrow", "wide", "other") NULL
```

this seems to fix it

deer-wmde (approved):

PR #914, opened 2025-04-09
feat: generate OAuth2 key pair dynamically within PHP

dati18:


app/WikiSetting.php
```diff
@@ -36,6 +36,8 @@ class WikiSetting extends Model
     public const wgSecretKey = 'wgSecretKey';
     public const wgLogo = 'wgLogo';
     public const wgFavicon = 'wgFavicon';
+    public const wgOAuth2PrivateKey = 'wgOAuth2PrivateKey';
+    public const wgOAuth2PublicKey = '$wgOAuth2PublicKey';
```
deer-wmde, line 40:
```suggestion
    public const wgOAuth2PublicKey = 'wgOAuth2PublicKey';
```

suggestion to make naming scheme the same as the other constants (without $)

deer-wmde:
I tested this successfully locally, both for new wikis and the existing ones (via the job). Great work!! 

super nitpick: proper CamelCase filename would have a big K I think: `app/Jobs/GenerateOAuth2KeysJob.php`

deer-wmde (approved):

deer-wmde:
https://phabricator.wikimedia.org/T336937

PR #916, opened 2025-04-16
Prevent Task overlaps

deer-wmde:
https://phabricator.wikimedia.org/T392055

https://laravel.com/docs/10.x/scheduling#preventing-task-overlaps

We suspect that at least `PollForMediaWikiJobsJob` is prone to multiple scheduling, so this could help


tarrow (commented):
Seems like we may as well try this to see if it does what we want

PR #917, opened 2025-04-16
add scheduler safeguard

deer-wmde:
alternative or addition to https://github.com/wbstack/api/pull/916

### DIFF
--- a/app/Console/Kernel.php
+++ b/app/Console/Kernel.php
@@ -29,6 +29,9 @@ class Kernel extends ConsoleKernel
      */
     protected function schedule(Schedule $schedule): void
     {
+        // do not schedule anything if this container is not "the scheduler"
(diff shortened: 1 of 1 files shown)


tarrow:
Did you already try this locally and confirm fewer jobs are fired?

deer-wmde:
> From my rough inspection of how scheduling tasks is supposed to work I think this is unnecessary and I would expect tasks from Kernel.php to only be scheduled by our scheduler pod.
> 
> Specifically I am looking at
> 
> https://github.com/wbstack/api/blob/e629b073596233d2518fb6374dc7c4f836ff7c64/start.sh#L26
> which appears to me to only schedule tasks on the scheduler pod anyway.
> 
> Therefore I think we should try #916 first in preference to this

ah! I think I missed this part.

PR #918, opened 2025-04-16
collect action numbers metrics

deer-wmde:
https://phabricator.wikimedia.org/T383427

based on https://github.com/wbstack/api/pull/912/

Adds data points to WikiMetrics model:
- daily_actions (number of actions in the last day)
- weekly_actions (number of actions in the last week)
- monthly_actions (number of actions in the last month)
- quarterly_actions (number of actions in the last quarter)

### DIFF
--- a/tests/Jobs/UpdateWikiDailyMetricJobTest.php
+++ b/tests/Jobs/UpdateWikiDailyMetricJobTest.php
@@ -39,12 +39,20 @@ class UpdateWikiDailyMetricJobTest extends TestCase
 
         $this->assertDatabaseHas('wiki_daily_metrics', [
             'wiki_id' => $activeWiki->id,
-            'date' => Carbon::today()->toDateString()
+            'date' => Carbon::today()->toDateString(),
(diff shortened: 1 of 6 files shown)


tests/Jobs/UpdateWikiDailyMetricJobTest.php
```diff
@@ -39,12 +39,20 @@ public function testRunJobForAllWikisIncludingDeletedWikis()
 
         $this->assertDatabaseHas('wiki_daily_metrics', [
             'wiki_id' => $activeWiki->id,
-            'date' => Carbon::today()->toDateString()
+            'date' => Carbon::today()->toDateString(),
```
tarrow, line 42: We talked about how we could "properly" test this and concluded that this is one of the main weaknesses of this approach. We're so tightly bound to the implementation / layout of mediawiki's private dbs that this isn't really possible to seed with good test data. Hence why this just lives as a null

tarrow (approved):

PR #919, opened 2025-04-17
Add number of triples count to the wiki daily metric table

rosalieper:


app/Metrics/App/WikiMetrics.php
```diff
@@ -14,15 +15,23 @@ public function saveMetrics(Wiki $wiki): void
...
             if ($oldRecord->is_deleted) {
-                \Log::info("Wiki is deleted, no new record for WikiMetrics ID {$wiki->id}.");
+                \Log::info("Wiki is deleted, no new record for Wiki ID {$wiki->id}.");
                 return;
             }
             if (!$isDeleted) {
                 if ($oldRecord->pages === $todayPageCount) {
-                    \Log::info("Page count unchanged for WikiMetrics ID {$wiki->id}, no new record added.");
-                    return;
+                    \Log::info("Page count unchanged for Wiki ID {$wiki->id}, no new record added.");
```
tarrow, line 30: oof, logic gets complicated here I think again but in this case a new record could still be added if the number of triples changes right?

I wonder if we now want to pull out all this logic into it's own method(s) to make it easier for humans to parse (and ensure we get the messages we send to users right)?

app/Metrics/App/WikiMetrics.php
```diff
@@ -14,15 +15,23 @@ public function saveMetrics(Wiki $wiki): void
...
+                    if ($oldRecord->number_of_triples === $tripleCount) {
+                        \Log::info("Number of tripples unchanged for Wiki ID {$wiki->id}, no new record added.");
```
tarrow, line 32:
```suggestion
                        \Log::info("Number of triples unchanged for Wiki ID {$wiki->id}, no new record added.");
```

tarrow:
Explodes for me locally when testing due to the TLS cert issues there.
`cURL error 60: SSL certificate problem: unable to get local issuer certificate (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) for https://asdfasdf.wbaas.dev/query/sparql?query=SELECT%20%28COUNT%28%2A%29%20AS%20%3Ftriples
  %29%20WHERE%20%7B%20%3Fs%20%3Fp%20%3Fo%20%7D`

At the moment I would think the best thing to do would be to import the certificates like we do for the tools (see: https://github.com/wbstack/quickstatements/commit/fe7ecd915c88b17a2669e2be0dd388c34a0350ac)

app/Metrics/App/WikiMetrics.php
```diff
@@ -2,9 +2,12 @@
 
 namespace App\Metrics\App;
 
+use App\QueryserviceNamespace;
 use App\Wiki;
 use App\WikiDailyMetrics;
 use Illuminate\Support\Arr;
+use Illuminate\Support\Facades\Http;
+use Illuminate\Support\Facades\Log;
```
tarrow, line 10: This feels nicer to me than `\Log` in line. Thanks!

tarrow (approved):

PR #931, opened 2025-05-06
 Extend wiki details with profile data

rosalieper:


rosalieper:
I did a rebase on main and updated the changelog file

PR #936, opened 2025-05-20
Add metrics: monthly casual users and active users

rosalieper:


app/WikiDailyMetrics.php
```diff
@@ -28,6 +28,8 @@ class WikiDailyMetrics extends Model
         'monthly_actions',
         'quarterly_actions',
         'number_of_triples',
+        'monthly_casual_users',
+        'monthly_active_users',
```
tarrow, line 32: This also needs to be added to the metricNames list

tarrow:
Mostly looks fine to me; as I mentioned yesterday let's get together and have another stab at adding some tests for this stuff because right now we don't which is a little unpleasant and makes touching these areas slower (we always need to manually test them.

deer-wmde:
I added the metric names in the static array, added bits to the migration code and tested the query on staging; If you don't see anything else I think this is ready for staging @tarrow

deer-wmde:
@rosalieper and I looked at this together and I stick the approve on it (even though the last changes were from me but we reviewd it)

deer-wmde (approved):

PR #937, opened 2025-05-22
Expand usage of LimitWikiAccess middleware

AndrewKostka:
Bug: [T394986](https://phabricator.wikimedia.org/T394986)

### DIFF
--- a/.github/workflows/docker.build.yml
+++ b/.github/workflows/docker.build.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v3
       - name: Set up Docker Buildx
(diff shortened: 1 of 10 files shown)


app/Http/Middleware/LimitWikiAccess.php
```diff
@@ -10,21 +10,32 @@
...
+        ])
+        ->with('wiki')
+        ->first();
 
-        if (!$userHasAccess) {
+        if (!$wikiManager) {
             abort(403);
         }
 
+        if (!$wikiManager->wiki) {
```
tarrow, line 34: I suspect this block should be higher in the middleware; for any random non-existing wiki I think this middleware currently returns a 403 not 404

PR #882, opened 2025-02-20
Add database script for MediaWiki 1.43

AndrewKostka:
Comparison of `mw1.39-wbs1` and `mw1.43-wbs1`:
https://www.diffchecker.com/rEJeYiDD/

PR #968, opened 2025-09-02
Update mw1.43-wbs1.sql

rosalieper:


rosalieper:
This change depends on https://github.com/wbstack/api/pull/882

PR #1082, opened 2026-04-11
Add `reuse_prototype` field to `/wiki` API endpoint

outdooracorn:
- add `reuse_prototype` field to `PublicWikiResource`
- add a dedicated `/reusePrototype` endpoint to decouple the prototype from our internal `/wiki` endpoint
- add unit tests for `PublicWikiController` behavior including test to ensure the additional `reuse_prototype` field doesn't cause an N+1 query issue
- add e2e test for `/reusePrototype` endpoint

Bug: T421877

### DIFF
--- a/app/Http/Controllers/PublicWikiController.php
+++ b/app/Http/Controllers/PublicWikiController.php
@@ -29,7 +29,7 @@ class PublicWikiController extends Controller {
         ]);
 
         $params = array_merge(self::$defaultParams, $request->input());
-        $query = Wiki::query();
+        $query = Wiki::query()->with('wikiLatestProfile');
 
         if (array_key_exists('is_featured', $params)) {
             $query = $query->where([
--- a/app/Http/Controllers/PublicWikiController.php
+++ b/app/Http/Controllers/PublicWikiController.php
@@ -72,6 +72,6 @@ class PublicWikiController extends Controller {
      * Display the specified resource.
      */
     public function show($id) {
-        return new PublicWikiResource(Wiki::findOrFail($id));
+        return new PublicWikiResource(Wiki::query()->with('wikiLatestProfile')->findOrFail($id));
     }
 }
--- a/app/Http/Controllers/PublicWikiController.php
+++ b/app/Http/Controllers/PublicWikiController.php
@@ -29,6 +29,7 @@ class PublicWikiController extends Controller {
         ]);
 
         $params = array_merge(self::$defaultParams, $request->input());
+        // eager load 'wikiLatestProfile' to avoid making N+1 queries
         $query = Wiki::query()->with('wikiLatestProfile');
 
         if (array_key_exists('is_featured', $params)) {
--- a/app/Http/Controllers/PublicWikiController.php
+++ b/app/Http/Controllers/PublicWikiController.php
@@ -72,6 +72,6 @@ class PublicWikiController extends Controller {
      * Display the specified resource.
      */
     public function show($id) {
-        return new PublicWikiResource(Wiki::query()->with('wikiLatestProfile')->findOrFail($id));
+        return new PublicWikiResource(Wiki::findOrFail($id));
     }
 }
--- a/app/Http/Resources/PublicWikiResource.php
+++ b/app/Http/Resources/PublicWikiResource.php
@@ -16,6 +16,21 @@ class PublicWikiResource extends JsonResource {
             'sitename' => $this->sitename,
             'wiki_site_stats' => $this->wikiSiteStats,
             'logo_url' => $logoSetting ? $logoSetting->value : null,
+
+            // TODO: delete these three fields before merging; here to easily prove the `reuse_prototype` logic works
+            'test_purpose' => $this->wikiLatestProfile ? $this->wikiLatestProfile->purpose : null,
+            'test_temporality' => $this->wikiLatestProfile ? $this->wikiLatestProfile->temporality : null,
+            'test_audience' => $this->wikiLatestProfile ? $this->wikiLatestProfile->audience : null,
+
+            // TODO: As the `$this->wikiLatestProfile` property can be accessed regardless of if
+            // `->with('wikiLatestProfile')` is used in the controller, we are unable to return null if
+            // `$this->wikiLatestProfile` isn't set. We should either look into addressing this, or remove the
+            // `$this->wikiLatestProfile ? ... : null` conditional.
+            'reuse_prototype' => $this->wikiLatestProfile
+                ? $this->wikiLatestProfile->purpose === 'data_hub'
+                  && $this->wikiLatestProfile->temporality === 'permanent'
+                  && $this->wikiLatestProfile->audience === 'wide'
+                : null,
         ];
     }
 }
--- a/app/Http/Resources/PublicWikiResource.php
+++ b/app/Http/Resources/PublicWikiResource.php
@@ -19,12 +19,10 @@ class PublicWikiResource extends JsonResource {
 
             // Checking relation load state before reading it to avoid N+1 query
             // This relies on the controller to eager load `wikiLatestProfile` relationship
-            'reuse_prototype' => $this->relationLoaded('wikiLatestProfile')
-                ? ($this->wikiLatestProfile
-                    ? $this->wikiLatestProfile->purpose === 'data_hub'
-                    && $this->wikiLatestProfile->temporality === 'permanent'
-                    && $this->wikiLatestProfile->audience === 'wide'
-                    : null)
+            'reuse_prototype' => $this->wikiLatestProfile
+                ? $this->wikiLatestProfile->purpose === 'data_hub'
+                  && $this->wikiLatestProfile->temporality === 'permanent'
+                  && $this->wikiLatestProfile->audience === 'wide'
                 : null,
         ];
     }
--- a/app/Http/Resources/PublicWikiResource.php
+++ b/app/Http/Resources/PublicWikiResource.php
@@ -17,11 +17,6 @@ class PublicWikiResource extends JsonResource {
             'wiki_site_stats' => $this->wikiSiteStats,
             'logo_url' => $logoSetting ? $logoSetting->value : null,
 
-            // TODO: delete these three fields before merging; here to easily prove the `reuse_prototype` logic works
-            'test_purpose' => $this->wikiLatestProfile ? $this->wikiLatestProfile->purpose : null,
-            'test_temporality' => $this->wikiLatestProfile ? $this->wikiLatestProfile->temporality : null,
-            'test_audience' => $this->wikiLatestProfile ? $this->wikiLatestProfile->audience : null,
-
             // Checking relation load state before reading it to avoid N+1 query
             // This relies on the controller to eager load `wikiLatestProfile` relationship
             'reuse_prototype' => $this->relationLoaded('wikiLatestProfile')
--- a/app/Http/Resources/PublicWikiResource.php
+++ b/app/Http/Resources/PublicWikiResource.php
@@ -22,14 +22,14 @@ class PublicWikiResource extends JsonResource {
             'test_temporality' => $this->wikiLatestProfile ? $this->wikiLatestProfile->temporality : null,
             'test_audience' => $this->wikiLatestProfile ? $this->wikiLatestProfile->audience : null,
 
-            // TODO: As the `$this->wikiLatestProfile` property can be accessed regardless of if
-            // `->with('wikiLatestProfile')` is used in the controller, we are unable to return null if
-            // `$this->wikiLatestProfile` isn't set. We should either look into addressing this, or remove the
-            // `$this->wikiLatestProfile ? ... : null` conditional.
-            'reuse_prototype' => $this->wikiLatestProfile
-                ? $this->wikiLatestProfile->purpose === 'data_hub'
-                  && $this->wikiLatestProfile->temporality === 'permanent'
-                  && $this->wikiLatestProfile->audience === 'wide'
+            // Checking relation load state before reading it to avoid N+1 query
+            // This relies on the controller to eager load `wikiLatestProfile` relationship
+            'reuse_prototype' => $this->relationLoaded('wikiLatestProfile')
+                ? ($this->wikiLatestProfile
+                    ? $this->wikiLatestProfile->purpose === 'data_hub'
+                    && $this->wikiLatestProfile->temporality === 'permanent'
+                    && $this->wikiLatestProfile->audience === 'wide'
+                    : null)
                 : null,
         ];
     }
--- a/app/Http/Resources/PublicWikiResource.php
+++ b/app/Http/Resources/PublicWikiResource.php
@@ -16,14 +16,11 @@ class PublicWikiResource extends JsonResource {
             'sitename' => $this->sitename,
             'wiki_site_stats' => $this->wikiSiteStats,
             'logo_url' => $logoSetting ? $logoSetting->value : null,
-
-            // Checking relation load state before reading it to avoid N+1 query
-            // This relies on the controller to eager load `wikiLatestProfile` relationship
             'reuse_prototype' => $this->wikiLatestProfile
                 ? $this->wikiLatestProfile->purpose === 'data_hub'
                   && $this->wikiLatestProfile->temporality === 'permanent'
                   && $this->wikiLatestProfile->audience === 'wide'
-                : null,
+                : false,
         ];
     }
 }
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -26,16 +26,5 @@ class AppServiceProvider extends ServiceProvider {
             $wrappedException = new \Exception("Executing Job '$name' failed.", 1, $event->exception);
             report($wrappedException);
         });
-
-        // Local-only SQL query logging for debugging
-        if ($this->app->environment('local')) {
-            \Event::listen(QueryExecuted::class, function (QueryExecuted $query) {
-                \Log::debug('Query Executed: ', [
-                    'sql' => $query->sql,
-                    'bindings' => $query->bindings,
-                    'connection' => $query->connectionName,
-                ]);
-            });
-        }
     }
 }
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -27,7 +27,7 @@ class AppServiceProvider extends ServiceProvider {
             report($wrappedException);
         });
 
-        // TODO: delete this listener before merging or is it useful to keep in the local environment?
+        // Local-only SQL query logging for debugging
         if ($this->app->environment('local')) {
             \Event::listen(QueryExecuted::class, function (QueryExecuted $query) {
                 \Log::debug('Query Executed: ', [
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -4,7 +4,6 @@ namespace App\Providers;
 
 use App\Http\Curl\CurlRequest;
 use App\Http\Curl\HttpRequest;
-use Illuminate\Database\Events\QueryExecuted;
 use Illuminate\Queue\Events\JobFailed;
 use Illuminate\Support\Facades\Queue;
 use Illuminate\Support\ServiceProvider;
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -4,6 +4,7 @@ namespace App\Providers;
 
 use App\Http\Curl\CurlRequest;
 use App\Http\Curl\HttpRequest;
+use Illuminate\Database\Events\QueryExecuted;
 use Illuminate\Queue\Events\JobFailed;
 use Illuminate\Support\Facades\Queue;
 use Illuminate\Support\ServiceProvider;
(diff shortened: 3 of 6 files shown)


ranyardm-wmde:
~~IMO given we don't have enough monitoring to predict how this will affect database load, this should ideally not be the accepted solution as it will create the multiple queries per search problem we discussed in ticket breakdown.~~  Apparently I'm wrong here.

app/Providers/AppServiceProvider.php
```diff
@@ -25,5 +26,16 @@ public function boot(): void {
             $wrappedException = new \Exception("Executing Job '$name' failed.", 1, $event->exception);
             report($wrappedException);
         });
+
+        // TODO: delete this listener before merging or is it useful to keep in the local environment?
```
dati18: I think it's useful and want to keep it here

rosalieper: I also find it helpful to keep it. Since we want to keep this we need to change the comment.

rosalieper (dismissed):
I am please with how this prototype did not turnout as a bigger change

rosalieper (dismissed):

app/Http/Resources/PublicWikiResource.php
```diff
@@ -16,6 +16,16 @@ public function toArray($request): array {
             'sitename' => $this->sitename,
             'wiki_site_stats' => $this->wikiSiteStats,
             'logo_url' => $logoSetting ? $logoSetting->value : null,
+
+            // Checking relation load state before reading it to avoid N+1 query
+            // This relies on the controller to eager load `wikiLatestProfile` relationship
+            'reuse_prototype' => $this->relationLoaded('wikiLatestProfile')
```
outdooracorn: This is good to think about to avoid accidentally creating an N+1 query issue in the future. I don't think it requires us to check this every time the `toArray()` method is called, though. Can we check that `PublicWikiController` eager loads the relationship in a test instead?

outdooracorn (commented):
From the [task's](https://phabricator.wikimedia.org/T421877) Task Breakdown section:

> * We would like to provide an alternative API endpoint so that the Suite team don't rely on an internal API endpoint that we could change at any point. We agreed to use `/reusePrototype` which points at the same `PublicWikiController` as the internal `/wiki` endpoint. See [routes/api.php#L59](https://github.com/wbstack/api/blob/0219964f59daefa1b7175472380db849486c56cb/routes/api.php#L59).
>    * If this turns out to be more than copy and editing [routes/api.php#L59](https://github.com/wbstack/api/blob/0219964f59daefa1b7175472380db849486c56cb/routes/api.php#L59) - 2hrs of work to write and review, then reconsider

I don't see any changes to `routes/api.php` or a comment saying that this was attempted but would take too long to implement.

@dati18 you were in the task breakdown meeting where we discussed this. How come it's not included in this PR?

@rosalieper remember, as the reviewer, it's important to read through the task so you know what the PR should contain.

app/Http/Resources/PublicWikiResource.php
outdooracorn: We might want to create a `PublicWikiControllerTest` unit test for this check. The current `PublicWikiTest` looks more like an e2e test which is not where I would expect this type of check to go. My hunch is that it will be easier to test in a unit test as well.

dati18: I somehow stashed the change and didn't notice. I also made a change to `show()` to make sure `wikiLatestProfile` was eager-loaded...
This is a stupid fault. I apologize.  :(

dati18: I added a unit test for the relation check

rosalieper:
> @rosalieper remember, as the reviewer, it's important to read through the task so you know what the PR should contain.

@outdooracorn Thank you for your comment, I read through the task description twice actually because it has a long description with updates and followup comment. Thanks for catching this. It left my mind at some point during the review but I will remember this incident. :)

app/Http/Resources/PublicWikiResource.php
dati18: based on your comment, I did the following:
- added the endpoint `/reusePrototype` and also made sure `show()` eager-loads `'wikiLatestProfile'`
- added PublicWikiControllerTest unit test
- modified the `reusePrototypeTest` by changing the endpoint to `reusePrototype` instead of generic `this->route`, also extended it to also call `GET /reusePrototype/id`

app/Http/Controllers/PublicWikiController.php
```diff
@@ -72,6 +72,6 @@ public function index(Request $request) {
      * Display the specified resource.
      */
     public function show($id) {
-        return new PublicWikiResource(Wiki::findOrFail($id));
+        return new PublicWikiResource(Wiki::query()->with('wikiLatestProfile')->findOrFail($id));
```
dati18: Same as `$query` in `index()`, I think we can benefit from eager loading here

tests/Http/Controllers/PublicWikiControllerTest.php
```diff
@@ -0,0 +1,34 @@
...
+            'purpose' => 'data_hub',
+            'temporality' => 'permanent',
+            'audience' => 'wide',
+        ]);
+
+        $controller = new PublicWikiController;
+        $resource = $controller->show($wiki->id);
+
+        $this->assertInstanceOf(PublicWikiResource::class, $resource);
+        $this->assertSame(true, $resource->toArray(request())['reuse_prototype']);
```
rosalieper: From the test name i understand we are testing eager loading but with this assertions i understand we are testing the `reuse_prototype` flag returns the corrects value which is good but i see no assertions for testing eager loading. Please let me know if you think i am missing something.

dati18: Thanks for pointing it out. I added a check if the model has the relation loaded.
I can also create a test queries counter and ensure 'wiki_profiles' is hit only once, but I think it's too much.

dati18: `relationLoaded('wikiLatestProfile')` returns true only if Laravel already loaded that relation before the resource tried to use it

app/Http/Controllers/PublicWikiController.php
outdooracorn: What makes you think that? Did you count the number of DB queries like I did in https://phabricator.wikimedia.org/T421877#11810168?

I'm wondering that as we are only getting a single `Wiki`, rather than a list of `Wiki`s, if the N+1 issue won't appear.

outdooracorn (commented):
Sorry, I didn't have enough time to provide a full review 🥲 

I do get this error when I checkout the PR, composer install, rebuild the container and then visit http://localhost:8082/wiki?sort=pages&direction=desc&page=1&per_page=10:

```
public/index.php:55
The stream or file "/var/www/html/storage/logs/laravel-2026-05-29.log" could not be opened in append mode: Failed to open stream: Permission denied The exception occurred while attempting to log: The stream or file "/var/www/html/storage/logs/laravel-2026-05-29.log" could not be opened in append mode: Failed to open stream: Permission denied The exception occurred while attempting to log: The stream or file "/var/www/html/storage/logs/laravel-2026-05-29.log" could not be opened in append mode: Failed to open stream: Permission denied The exception occurred while attempting to log: The stream or file "/var/www/html/storage/logs/laravel-2026-05-29.log" could not be opened in append mode: Failed to open stream: Permission denied The exception occurred while attempting to log: The stream or file "/var/www/html/storage/logs/laravel-2026-05-29.log" could not be opened in append mode: Failed to open stream: Permission denied The exception occurred while attempting to log: The stream or file "/var/www/html/storage/logs/laravel-2026-05-29.log" could not be opened in append mode: Failed to open stream: Permission denied The exception occurred while attempting to log: Query Executed: Context: {"sql":"select count(*) as aggregate from `wikis` where `wikis`.`deleted_at` is null","bindings":[],"connection":"mysql"} Context: {"exception":{}} Context: {"exception":{}} Context: {"exception":{}} Context: {"exception":{}} Context: {"exception":{}} 
```

This doesn't happen when I checkout `origin/main`, composer install, rebuild the container, and then visit http://localhost:8082/wiki?sort=pages&direction=desc&page=1&per_page=10. So there is likely something for you to look into here. Hopefully I haven't done anything silly locally! 🙈

app/Http/Controllers/PublicWikiController.php
dati18: > I do get this error when I checkout the PR, composer install, rebuild the container and then visit

Can you pasted the command? I just run composer install and didn't see similar error. I tore down everything and rebuilt/reinstalled also...

dati18: > I'm wondering that as we are only getting a single Wiki, rather than a list of Wikis, if the N+1 issue won't appear.

Ah you're right, it got me right there... since `show()` returns a single wiki, worst-case scenario is `wikiLatestProfile` was lazy-loaded and we do 1 extra query.

So I have 2 options:
- Keep it for the sake of explicitness/consistency and avoid hidden lazy loading.
- Reverse it back because its impact is very much minimal

What do you think?

dati18: @rosalieper your opinion is also valuable as well :)

tests/Http/Controllers/PublicWikiControllerTest.php
```diff
@@ -0,0 +1,34 @@
...
+use App\Http\Resources\PublicWikiResource;
+use App\Wiki;
+use App\WikiProfile;
+use Illuminate\Foundation\Testing\DatabaseTransactions;
+use Tests\TestCase;
+
+class PublicWikiControllerTest extends TestCase {
+    use DatabaseTransactions;
+
+    public function testShowEagerLoadsWikiLatestProfileForResource(): void {
```
outdooracorn: ~~This test passes even if `wikiLatestProfile` isn't eager loaded.~~

~~The `PublicWikiResource` will lazy load the model if it hasn't already been loaded so `$this->assertTrue($resource->resource->relationLoaded('wikiLatestProfile'));` will always be true.~~

~~Looking at `PublicWikiController` a bit closer, it might be harder than I initially thought to test this.~~

EDIT: the above is incorrect.

[You mentioned in another comment](https://github.com/wbstack/api/pull/1082#discussion_r3319115908) about creating a test queries counter and ensure 'wiki_profiles' is hit only once. Is this something that Laravel provides helpers for? This _might_ be useful if not too difficult to implement.

app/Http/Resources/PublicWikiResource.php
outdooracorn: If we can create a successful test to check for eager loading, then we can remove this check here.

outdooracorn (commented):
I tested again locally and couldn't reproduce my previous error, so not sure what the issue was there. Sorry for the noise.

tests/Http/Controllers/PublicWikiControllerTest.php
dati18: Are you sure? 
If I change `show()` to lazy load the test fails... :(

```php
public function show($id) {
        return new PublicWikiResource(Wiki::query()->findOrFail($id));
}
```
```php
public function show($id) {
        return new PublicWikiResource(Wiki::findOrFail($id));
}
```

dati18: @outdooracorn can you test it again?

outdooracorn: Oh! I see what happened here, I removed the eager loading from the `index()` method as that is the method that **_requires_** eager loading to avoid N+1 issues. I hadn't noticed you wrote this test only for the `show()` method. Why aren't testing the `index()` method?

app/Providers/AppServiceProvider.php
```diff
@@ -25,5 +26,16 @@ public function boot(): void {
...
+        // Local-only SQL query logging for debugging
+        if ($this->app->environment('local')) {
+            \Event::listen(QueryExecuted::class, function (QueryExecuted $query) {
+                \Log::debug('Query Executed: ', [
+                    'sql' => $query->sql,
+                    'bindings' => $query->bindings,
+                    'connection' => $query->connectionName,
+                ]);
+            });
+        }
```
outdooracorn: `The stream or file "/var/www/html/storage/logs/laravel-2026-05-29.log" could not be opened in append mode: Failed to open stream: Permission denied`

I got the above error again and tracked it down to:
* some of the logs in `/var/www/html/storage/logs/` were owned by `root:root` rather than `www-data:www-data`
* this event listening doing a log

I don't think it's an issue in staging/production as I don't think logging to a file is enabled.

dati18: ah I just checked and had the same issue. But you've already specified "local"-only so I don't think it's going to be a problem on staging/prod

tests/Http/Controllers/PublicWikiControllerTest.php
dati18: Ah yes... `index()` returns a collection of wikis so it's worth testing than `show()`
I just added a test for `index()` by by generating a random `n` number of wikis (from 3 to 100) and checking for the number of queries (it should be 1) and making sure `wikiLatestProfile` is eager-loaded.

dati18: fun read about relation loaded and stuff 

https://laravel-news.com/relationloaded

tests/Http/Controllers/PublicWikiControllerTest.php
```diff
@@ -0,0 +1,67 @@
...
+
+        $profileQueryCount = 0;
+        DB::listen(function (QueryExecuted $query) use (&$profileQueryCount): void {
+            if (str_contains($query->sql, 'wiki_profiles')) {
+                $profileQueryCount++;
+            }
+        });
+
+        $controller = new PublicWikiController;
+        $resourceCollection = $controller->index(request());
```
outdooracorn: The `request()` method this for getting "an instance of the current request or an input item from the request" which does feel quite right to use here. I'd prefer to create an empty `Request` instance instead.

```suggestion
        $resourceCollection = $controller->index(new Request());
```

tests/Http/Controllers/PublicWikiControllerTest.php
```diff
@@ -0,0 +1,67 @@
...
+
+        $this->assertSame(1, $profileQueryCount);
+        $this->assertTrue($resourceCollection->collection[0]->resource->relationLoaded('wikiLatestProfile'));
```
outdooracorn: Optional nitpick: if you want, you can use `first()` instead of `collection[0]` and `->resource` isn't required.

```suggestion
        $this->assertTrue($resourceCollection->first()->relationLoaded('wikiLatestProfile'));
```

tests/Http/Controllers/PublicWikiControllerTest.php
outdooracorn: Cool that you have figured out how to count the SQL queries executed.

However, there are some things I don't like about this new test:
* Tests should be deterministic - given the same code and environment, they should behave the same way every time they run. Creating a random number of Wikis is not deterministic. I would create 3 - 5 wikis for this test.  
* Tests should not `echo` or otherwise print output. Instead, they should communicate results through assertions, exceptions, and PHPUnit's output mechanisms

tests/Http/Controllers/PublicWikiControllerTest.php
```diff
@@ -0,0 +1,67 @@
...
+        $this->assertTrue($resource->resource->relationLoaded('wikiLatestProfile'));
+    }
+
+    public function testIndexEagerLoadsWikiLatestProfileOnceForCollection(): void {
+        $n = rand(3, 100);
+        echo $n;
+        for ($i = 1; $i <= $n; $i++) {
+            $wiki = Wiki::factory()->create([
+                'domain' => 'index-eager-load-test-' . $i . '.wikibase.cloud',
+                'sitename' => 'Index Eager Load Test Site ' . $i,
```
outdooracorn: Nitpick: We are using a PHP version that supports [string interpolation](https://www.php.net/manual/en/language.types.string.php#language.types.string.parsing) so lets use it as it reduces the line length. I also find it much easier to read.

```suggestion
                'domain' => "index-eager-load-test-{$i}.wikibase.cloud",
                'sitename' => "Index Eager Load Test Site {$i}",
```

app/Http/Controllers/PublicWikiController.php
outdooracorn: My vote is to remove it from this PR as:
* it is not required
* it is not related to the task ([T421877](https://phabricator.wikimedia.org/T421877))

It's good to keep PRs small and to the point where possible.

outdooracorn (commented):
As requested by @dati18, not a full review. Mostly comments around the `PublicWikiControllerTest`.

app/Http/Resources/PublicWikiResource.php
```diff
@@ -16,6 +16,14 @@ public function toArray($request): array {
             'sitename' => $this->sitename,
             'wiki_site_stats' => $this->wikiSiteStats,
             'logo_url' => $logoSetting ? $logoSetting->value : null,
+
+            // Checking relation load state before reading it to avoid N+1 query
+            // This relies on the controller to eager load `wikiLatestProfile` relationship
```
outdooracorn: This comment is incorrect now that the `->relationLoaded()` check has been moved to a test.

I would remove this comment. If we want to mention eager loading in the code then add a comment above PublicWikiController.php#L32 as that is where the eager loading is happening.

app/Http/Resources/PublicWikiResource.php
outdooracorn: Looking at the Acceptance Criteria in the task, this should be `false` rather than `null`.

> * [] The flag is 1/true if the instance has:
>   * purpose = data_hub AND
>   * lifespan = permanent AND
>   * audience = wide
> * [] The flag is 0/false otherwise.

app/Providers/AppServiceProvider.php
outdooracorn: I think remove. We can add it when we want to use it. Not useful for day-to-day debugging.

routes/api.php
```diff
@@ -57,5 +57,6 @@
     });
 
     $router->apiResource('wiki', 'PublicWikiController')->only(['index', 'show']);
+    $router->apiResource('reusePrototype', 'PublicWikiController')->only(['index', 'show']);
```
outdooracorn: I'm thinking we should just enable the `index` method for the reuse prototype as that is all that is requested in the task.

tests/Http/Controllers/PublicWikiControllerTest.php
```diff
@@ -0,0 +1,66 @@
...
+
+            WikiProfile::create([
+                'wiki_id' => $wiki->id,
+                'purpose' => 'data_hub',
+                'temporality' => 'permanent',
+                'audience' => 'wide',
+            ]);
+        }
+
+        $profileQueryCount = 0;
```
outdooracorn: Nitpick: `profile` makes me think of "profiling", which this is kind of doing, but I don't think that is what the variable name is meaning. It is the counter for queries that contain `wiki_profiles`.


```suggestion
        $wikiProfilesQueryCount = 0;
```

tests/Http/Controllers/PublicWikiControllerTest.php
```diff
@@ -0,0 +1,66 @@
...
+        ]);
+
+        $controller = new PublicWikiController;
+        $resource = $controller->show($wiki->id);
+
+        $this->assertInstanceOf(PublicWikiResource::class, $resource);
+        $this->assertSame(true, $resource->toArray(new Request)['reuse_prototype']);
+    }
+
+    public function testIndexEagerLoadsWikiLatestProfileOnceForCollection(): void {
```
outdooracorn: Let's add a test that checks `false` is returned when there is no `WikiProfile` information in the database.

outdooracorn (commented):
Sorry, didn't manage a full review, I have to run away now.

routes/api.php
dati18: I did it and it means I also had to update `testReusePrototype()`, and we can't load a single wiki anymore aka `reusePrototype/{$id}` is no longer a valid route just so you know

app/Http/Resources/PublicWikiResource.php
dati18: Done. Summary:
- "non-reusable" wiki still returns `false`
- "no-profile" wiki returns `false` instead of `null`

tests/Http/Controllers/PublicWikiControllerTest.php
dati18: 2 tests added:
- return `false` when a wiki doesn't match reuse prototype
- return `false` when a wiki doesn't have a `WikiProfile`

outdooracorn: Thanks!

tests/Http/Controllers/PublicWikiControllerTest.php
outdooracorn: Have refactored this to use `$wikiProfilesQueryCount` (plural) to match `wiki_profiles` (plural)

routes/api.php
outdooracorn: > we can't load a single wiki anymore aka `reusePrototype/{$id}` is no longer a valid route just so you know

Correct, that is what I'm requesting, because it's not part of the task.

dati18 (dismissed):

dati18 (approved):

PR #1096, opened 2026-04-14
Allow dependabot composer "direct" updates

tarrow:
We hope that this includes all production and dev dependencies but we need to confirm by doing because the documentation isn't clear.


outdooracorn (approved):
I did some digging and found a JSON Schema to validate this change against. I believe this is the correct syntax.

PR #1105, opened 2026-04-21
Add tests for HorizonPrometheusExporter and our Custom Exporters

tarrow:
Add some simple tests. These instantiate the existing exporters and check that their collection method works. Another test instantiates the exporter repository with our application config to check it is valid

Bug: T423429

### DIFF
--- a/tests/HorizonPrometheusExporter/ExporterRepositoryTest.php
+++ b/tests/HorizonPrometheusExporter/ExporterRepositoryTest.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace Tests\HorizonPrometheusExporter;
+
+use LKDevelopment\HorizonPrometheusExporter\Repository\ExporterRepository;
+use Prometheus\CollectorRegistry;
+use Tests\TestCase;
+
+class ExporterRepositoryTest extends TestCase {
+    public function testCanBeCreated() {
+        // if nothing is passed to load it reads the config
+        ExporterRepository::load();
+        $registry = ExporterRepository::getRegistry();
+        $this->assertInstanceOf(CollectorRegistry::class, $registry);
+    }
+}
--- a/tests/HorizonPrometheusExporter/ExporterRepositoryTest.php
+++ b/tests/HorizonPrometheusExporter/ExporterRepositoryTest.php
@@ -8,7 +8,7 @@ use Tests\TestCase;
 
 class ExporterRepositoryTest extends TestCase {
     public function testCanBeCreated() {
-        // if nothing is passed to load it reads the config
+        // if nothing is passed to load(), exporters are read from the config at config/horizon-exporter.php
         ExporterRepository::load();
         $registry = ExporterRepository::getRegistry();
         $this->assertInstanceOf(CollectorRegistry::class, $registry);
--- a/tests/Metrics/FailedQsBatchesTest.php
+++ b/tests/Metrics/FailedQsBatchesTest.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace Tests\Metrics;
+
+use App\Metrics\FailedQsBatches;
+use LKDevelopment\HorizonPrometheusExporter\Repository\ExporterRepository;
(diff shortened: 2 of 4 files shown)


tests/HorizonPrometheusExporter/ExporterRepositoryTest.php
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Tests\HorizonPrometheusExporter;
+
+use LKDevelopment\HorizonPrometheusExporter\Repository\ExporterRepository;
+use Prometheus\CollectorRegistry;
+use Tests\TestCase;
+
+class ExporterRepositoryTest extends TestCase {
+    public function testCanBeCreated() {
+        // if nothing is passed to load it reads the config
```
outdooracorn:
```suggestion
        // if nothing is passed to load(), exporters are read from the config at config/horizon-exporter.php
```

tests/Metrics/FailedQsBatchesTest.php
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Tests\Metrics;
+
+use App\Metrics\FailedQsBatches;
+use LKDevelopment\HorizonPrometheusExporter\Repository\ExporterRepository;
+use Tests\TestCase;
+
+class FailedQsBatchesTest extends TestCase {
+    public function testCanBeCreated() {
+        ExporterRepository::load([FailedQsBatches::class]);
```
outdooracorn, line 11: How about we load this from the config file as well so that we are testing that as well?

```suggestion
        // if nothing is passed to load(), exporters are read from the config at config/horizon-exporter.php
        ExporterRepository::load();
```

tests/Metrics/PendingQsBatchesTest.php
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Tests\Metrics;
+
+use App\Metrics\PendingQsBatches;
+use LKDevelopment\HorizonPrometheusExporter\Repository\ExporterRepository;
+use Tests\TestCase;
+
+class PendingQsBatchesTest extends TestCase {
+    public function testCanBeCreated() {
+        ExporterRepository::load([PendingQsBatches::class]);
```
outdooracorn, line 11: How about we load this from the config file as well so that we are testing that as well?

```suggestion
        // if nothing is passed to load(), exporters are read from the config at config/horizon-exporter.php
        ExporterRepository::load();
```

tests/Metrics/WikiEntityImportsTest.php
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Tests\Metrics;
+
+use App\Metrics\WikiEntityImports;
+use LKDevelopment\HorizonPrometheusExporter\Repository\ExporterRepository;
+use Tests\TestCase;
+
+class WikiEntityImportsTest extends TestCase {
+    public function testCanBeCreated() {
+        ExporterRepository::load([WikiEntityImports::class]);
```
outdooracorn, line 11: How about we load this from the config file as well so that we are testing that as well?

```suggestion
        // if nothing is passed to load(), exporters are read from the config at config/horizon-exporter.php
        ExporterRepository::load();
```

outdooracorn (approved):
Thanks for this; it looks good!

I've left some minor comments - take them or leave them :)

tests/Metrics/FailedQsBatchesTest.php
tarrow: As we just discussed that will then make this test fragile to loading *any* of our exporters. Let's consider adding a follow up patch to check it in the ExporterRepositoryTest but not here

tests/Metrics/PendingQsBatchesTest.php
tarrow: as above

tests/Metrics/WikiEntityImportsTest.php
tarrow: as above

PR #1132, opened 2026-05-19
Add Carbon 3 boundary coverage for timestamp parsing and integer-cast diff behavior

dati18:
- `MWTimestampHelperTest.php`: Removed `if` block that checked for type of `$carbon` since `getCarbonFromMWTimestamp()` returns `CarbonImmutable`, `TypeError` will be thrown if not.
- `SendEmptyWikiNotificationsJobTest.php`: Tightened notification-threshold coverage so a wiki that is almost old enough (for example: 29.5 days old) does not trigger an empty-wiki notification early.
- `PlatformStatsSummaryJobTest.php`: Added a test to ensure a second-precision `lastEdit` timestamp exactly at the inactivity cutoff is still counted as active.
- `ConversionMetricTest.php`: Added a test to verify fractional day diffs are truncated to the integer values expected by the conversion metrics API.

Bug: T426592


app/Helper/MWTimestampHelper.php
```diff
@@ -15,8 +15,13 @@ class MWTimestampHelper {
     private const MWTimestampFormat = 'YmdHis';
 
     public static function getCarbonFromMWTimestamp(string $MWTimestamp): CarbonImmutable {
-        $carbon = CarbonImmutable::createFromFormat(self::MWTimestampFormat, $MWTimestamp);
-        if ($carbon === null) {
+        try {
```
deer-wmde: I'm a bit unclear about this part.

1. since Carbon 3 will throw an exception when `createFormFormat`, I'm not certain it helps to catch it just so we can throw our own. This could even make it harder to spot the reason for why it failed. 
2. the second check also seems obsolete since we either get a `CarbonImmutable` or an Exception gets thrown

So I think this method could be as simple as just line 19 - if something goes wrong an exception gets thrown as it used to. Is there more context for this additional logic that I'm not aware of?

tests/Helper/MWTimestampHelperTest.php
```diff
@@ -19,6 +19,7 @@ public function testGetCarbonFromMWTimestamp() {
 
     public function testGetCarbonFromMWTimestampWithInvalidTimestamp() {
         $this->expectException(InvalidFormatException::class);
+        $this->expectExceptionMessage('Unable to create Carbon object');
```
deer-wmde: As line 21 is expecting the Exception already, I struggle to see how this check helps us

app/Helper/MWTimestampHelper.php
dati18: you're right to question it, it's an overthinking part from me. My logic is `createFromFormat()` may throw a non-instance (`null`/`false`), so an extra `instanceof` guard prevents type error when returning `$carbon`.

tests/Helper/MWTimestampHelperTest.php
dati18: I just want to lock down the exact error message. a bit redundant, yes... WHat do you think? keep it or remove it?

app/Helper/MWTimestampHelper.php
deer-wmde: Ah true, I didn't realize the returned value could also be `null`

tests/Helper/MWTimestampHelperTest.php
deer-wmde: yes, this thought was connected to the other one. If we keep throwing the Exception ourselves I think it's fine to keep it, might even be worth to set the message to something that explicitly makes it obvious it comes from "us" and not just the library

app/Helper/MWTimestampHelper.php
outdooracorn: @dati18 I like your defensive programming thinking here.

Your reply explains why the `=== null`  check was replaced with an `instaceof` check; but how come the catching and re-throwing of the same exception? 

I agree with @deer-wmde here in thinking that allowing the `InvalidFormatException` to be thrown from `createFromFormat()` is enough. I'm thinking about the [KISS principle](https://en.wikipedia.org/wiki/KISS_principle).

tests/Helper/MWTimestampHelperTest.php
outdooracorn: I also tend to shy away from validating error messages unless it is some sort of product requirement or is somehow crucial to the function of the unit under test. I find that in these situations you often end up creating more work for yourself (updating both class and test) without making the program/test suite any more robust.

app/Helper/MWTimestampHelper.php
```diff
@@ -15,8 +15,13 @@ class MWTimestampHelper {
...
+            $carbon = CarbonImmutable::createFromFormat(self::MWTimestampFormat, $MWTimestamp);
+        } catch (InvalidFormatException $exception) {
+            throw new InvalidFormatException('Unable to create Carbon object', 0, $exception);
```
outdooracorn: Is this actually an Invalid Format Exception? I would assume that if it was one of those then `CarbonImmutable::createFromFormat()` would throw it. At a glance, this seems more like a [`RuntimeException`](https://www.php.net/manual/en/class.runtimeexception.php) to me.

app/Helper/MWTimestampHelper.php
outdooracorn: Thinking out loud some more, if `CarbonImmutable::createFromFormat()` only returns `Carbon` or `null` then `if ($carbon === null)` should be equivalent to `if (!$carbon instanceof CarbonImmutable)`.

However, given that my PHP knowledge isn't good enough to be 100% confident that `createFromFormat()` won't return anything other than `Carbon` or `null` without throwing an error, and we haven't currently identified this as part of the codebase that needs optimizing, I'm happy to keep using the more explicit `instaceof` check. 🙂

app/Helper/MWTimestampHelper.php
dati18: Good catch! I was thinking about how to answer this question. My knowledge is that `InvalidFormatException` is good for catching a single exception type for all "I can't parse MW timestamp" failures. A different type exception (like `RuntimeException` or `UnexpectedValueException`) is better if you want to distinguish what kind of exception was caught. 
I think using `InvalidFormatException` is good enough; both cases check for input-parsing failures.

app/Helper/MWTimestampHelper.php
dati18: I think `instanceof` is more robust against unexpected values and a safer overall check if behavior changes later. So we keep it?

dati18: I also updated the exception messages to be clearer and more precise.

tests/Helper/MWTimestampHelperTest.php
outdooracorn, line 1: This test suite doesn't fail when `MWTimestampHelper` hasn't been modified, so this PR doesn't currently guard against an issue that #1130 was created to prevent.

tests/Jobs/SendEmptyWikiNotificationsJobTest.php
outdooracorn, line 1: This test suite doesn't fail when `SendEmptyWikiNotificationsJob` hasn't been modified, so this PR doesn't currently guard against an issue that #1130 was created to prevent.

outdooracorn (changes requested):
In order to verify that these tests guard against the issues addressed in #1130, I did the following:
* Checkout out the main branch just before #1130 was merged at commit 02b92c3
* Did a `composer install` (using the Docker Compose command in the README)
* Merged in the changes to the tests in this PR
* Commented out [`MWTimestampHelperTest:22`](https://github.com/wbstack/api/pull/1132/changes#diff-6ce47f8e1888ecf7459eb7a4f891707a648a74e5693faa8dc87875ada4ce77d9R22) as that isn't really a guard against the issues addressed in #1130
* Ran the tests in this PR with `dc exec -it api vendor/bin/phpunit tests/Helper/MWTimestampHelperTest.php tests/Jobs/PlatformStatsSummaryJobTest.php tests/Jobs/SendEmptyWikiNotificationsJobTest.php tests/Routes/Wiki/ConversionMetricTest.php`

If all issues addressed in #1130 now have a test to guard against that, I would expect at least one method in the 4 test classes to fail. Instead, only tests in `PlatformStatsSummaryJobTest`  and `ConversionMetricTest` failed meaning we are currently missing some test coverage.

app/Helper/MWTimestampHelper.php
outdooracorn: Yes, I'm happy to keep the `instaceof` check. Let's keep discussing what sort of exception should be thrown in [the other comment thread](https://github.com/wbstack/api/pull/1132#discussion_r3304317751), though.

---

I still think that we don't need the `try`/`catch` block though. This extra block isn't required for the functionality of this helper or how it is used in `PlatformStatsSummaryJob`, so I think we should keep the code simple and less verbose. Something like:

```

$carbon = CarbonImmutable::createFromFormat(self::MWTimestampFormat, $MWTimestamp);
if (!$carbon instanceof CarbonImmutable) {
    // throw runtime or logic exception
}

return $carbon
```

Do you have any thoughts on that?

tests/Helper/MWTimestampHelperTest.php
dati18: `MWTimestampHelperTest` only hits the code paths where `createFromFormat()` never returns `false`.

- In `testGetCarbonFromMWTimestampWithInvalidTimestamp()`, the [test](https://github.com/wbstack/api/blob/0219964f59daefa1b7175472380db849486c56cb/tests/Helper/MWTimestampHelperTest.php#L21) only expects _some exception_ for invalid input. In our environment, `CarbonImmutable::createFromFormat()` itself throws `InvalidFormatException` before execution ever reaches `if ($carbon === false) { throw new \Exception('Unable to create Carbon object'); }`, the test can still pass because an exception was thrown anw.
- In `testGetCarbonFromMWTimestampWithUnixTimestamp()`, same concept, [`InvalidFormatException` was thrown first](https://github.com/wbstack/api/blob/0219964f59daefa1b7175472380db849486c56cb/tests/Helper/MWTimestampHelperTest.php#L28), `=== false` was never reached

Basically the test did not prove this older assumption: invalid input returns `false`. This assumption is where Carbon 3 changed behavior, so the bug is hidden.

app/Helper/MWTimestampHelper.php
outdooracorn: > I think using `InvalidFormatException` is good enough; both cases check for input-parsing failures.

Do we know for sure that `CarbonImmutable::createFromFormat()`  will only return something that is not an instance of `CarbonImmutable` when there is an input-parsing failure? Said another way, could there be other types of failures that result in something that is not an instance of `CarbonImmutable` being returned? 

I think the answer to that is "we don't know for sure". Therefore, we shouldn't throw an `InvalidFormatException` if it might not actually be an invalid format that caused unexpected return type.

It also doesn't matter too much, for the functionality of `MWTimestampHelper` or how it is used in `PlatformStatsSummaryJob`, what the exception is (as long as it's not potentially incorrect) or whether we want to be able to catch a single exception [1] vs distinguish between multiple exceptions. What matters, at least to me, is that we fail early when we notice an unexpected return value, and we don't propagate that unexpected value to other parts of the stack which could have unexpected and harder to debug consequences.

I therefore still think that we should use either:
* a [`LogicException`](https://www.php.net/manual/en/class.logicexception.php) if we think that this is an error in the program logic which should lead directly to a fix in our code, or 
* a [`RuntimeException`](https://www.php.net/manual/en/class.runtimeexception.php) if we think that this is an error which can only be found at runtime.

---

[1]:  Side note - if a user of the `MWTimestampHelper` wanted to be able to catch a single exception then using the same but potentially incorrect exception isn't be the way I would address that requirement.

tests/Jobs/SendEmptyWikiNotificationsJobTest.php
dati18: The function `checkIfWikiIsOldAndEmpty()` avoids the Carbon 3 sneakily by using a non-negative diff `$emptyWikiDays = $createdAt->diffInDays($now, true)` and comparing it with an integer `$emptyDaysThreshold`  (it's 30 days). For example the wiki age is `29.99` days is still considered < 30 and the comparison is `false`.
In this specific method, Carbon 3 float change doesn't change the behavior. test was still green.

dati18: casting `$emptyWikiDays` with `(int)` just strengthens the Carbon 3 behavior change.

app/Helper/MWTimestampHelper.php
dati18: For the sake of [KISS](https://en.wikipedia.org/wiki/KISS_principle), I think I prefer to keep it simple. 
updated

tests/Helper/MWTimestampHelperTest.php
dati18: removed.

app/Helper/MWTimestampHelper.php
dati18: since it doesn't affect functionality of the method, I removed the `catch` and keep it simple

tests/Helper/MWTimestampHelperTest.php
outdooracorn: > `MWTimestampHelperTest` only hits the code paths where `createFromFormat()` never returns `false`.

That seems like the test needs improving then!

> * In `testGetCarbonFromMWTimestampWithInvalidTimestamp()`, the [test](https://github.com/wbstack/api/blob/0219964f59daefa1b7175472380db849486c56cb/tests/Helper/MWTimestampHelperTest.php#L21) only expects _some exception_ for invalid input. 

No, the test that you linked to expects that an exception that is an instance of `InvalidFormatException` is thrown.

> Basically the test did not prove this older assumption: invalid input returns `false`. This assumption is where Carbon 3 changed behavior, so the bug is hidden.

I'm not sure that we can say "invalid input returns `false`". Even in older versions of Carbon, the default is to throw an exception rather than return `false` or `null`.

---

Looking into this some more and Carbon has a [strict mode](https://carbon.nesbot.com/develop/reference.html#carbon-usestrictmode). When this is `true`, which is the default, then an `InvalidFormatException` is thrown. When it is `false` then `false` or `null` is returned.

We could have a test that sets strict mode to `false` in order to test this `if (!$carbon instanceof CarbonImmutable)` conditional. However, I think we can actually remove this conditional altogether without affecting this functionality. `MWTimestampHelper::getCarbonFromMWTimestamp()` requires that a `CarbonImmutable` is returned. If `false` or `null` or anything that isn't an instance of `CarbonImmutable` is returned then a `TypeError` is thrown. `getCarbonFromMWTimestamp()` will either return a `CarbonImmutable` or throw an exception.

tests/Routes/Wiki/ConversionMetricTest.php
```diff
@@ -58,10 +58,10 @@ private function createTestWiki($name, $createdWeeksAgo, $firstEditedWeeksAgo, $
         $wiki->created_at = $current_date->subWeeks($createdWeeksAgo);
         $events = $wiki->wikiLifecycleEvents();
         $update = [];
-        if ($lastEditedWeeksAgo) {
+        if ($lastEditedWeeksAgo !== null) {
             $update['last_edited'] = $current_date->subWeeks($lastEditedWeeksAgo);
         }
-        if ($firstEditedWeeksAgo) {
+        if ($firstEditedWeeksAgo !== null) {
```
outdooracorn: Why are these changes required? This `createTestWiki()` method is not used in the new `testDownloadJsonTruncatesFractionalDayDiffs()` method you've added.

deer-wmde (dismissed):

app/Jobs/SendEmptyWikiNotificationsJob.php
```diff
@@ -34,6 +34,7 @@ public function checkIfWikiIsOldAndEmpty(Wiki $wiki) {
         $emptyDaysThreshold = config('wbstack.wiki_empty_notification_threshold');
         $createdAt = $wiki->created_at;
         $now = CarbonImmutable::now();
+        // Normalize to whole days so comparisons with the integer config threshold stay deterministic.
```
outdooracorn: The Carbon 3 change to return `float` instead of `int` doesn't make this non-deterministic. How about this comment that includes some context about the previous Carbon 2 behavior?

```suggestion
        // cast to int to retain Carbon 2 behaviour of using whole days
```

app/Http/Controllers/ConversionMetricController.php
```diff
@@ -34,13 +34,16 @@ public function index(Request $request) {
             $daysSinceLastEdit = null;
 
             if ($wikiLastEditedTime !== null) {
+                // Cast to int to avoid fractional-day values in CSV/JSON and keep cutoff logic predictable.
```
outdooracorn: Technically the cutoff logic is predictable. How about this comment that includes some context about the previous Carbon 2 behavior?

```suggestion
                // cast to int to retain Carbon 2 behaviour of using whole days
```

app/Http/Controllers/ConversionMetricController.php
```diff
@@ -34,13 +34,16 @@ public function index(Request $request) {
...
                 $daysSinceLastEdit = (int) $wikiLastEditedTime->diffInDays($current_date, false);
             }
 
             if ($daysSinceLastEdit !== null && $daysSinceLastEdit >= 90) {
+                // Keep exported duration as whole days for stable reporting.
```
outdooracorn: I'm not sure how whole vs fractional days affects the reporting stability. How about this comment that includes some context about the previous Carbon 2 behavior?

```suggestion
                // cast to int to retain Carbon 2 behaviour of using whole days
```

app/Http/Controllers/ConversionMetricController.php
```diff
@@ -34,13 +34,16 @@ public function index(Request $request) {
...
                 $time_before_wiki_abandoned_days = (int) $wiki->created_at->diffInDays($wikiLastEditedTime, false);
             }
             if ($wikiFirstEditedTime !== null) {
+                // Keep exported duration as whole days for stable reporting.
```
outdooracorn: I'm not sure how whole vs fractional days affects the reporting stability. How about this comment that includes some context about the previous Carbon 2 behavior?

```suggestion
                // cast to int to retain Carbon 2 behaviour of using whole days
```

app/Jobs/PlatformStatsSummaryJob.php
```diff
@@ -132,6 +132,7 @@ public function prepareStats(array $allStats, $wikis): array {
             // is it edited in the last 90 days?
             if (!is_null($stats['lastEdit'])) {
                 $lastTimestamp = MWTimestampHelper::getCarbonFromMWTimestamp(intval($stats['lastEdit']));
+                // Carbon can return fractional seconds; cast to int so threshold checks use whole-second precision.
```
outdooracorn: The `$absolute = true` argument was also added. How about these comments that include some context about the previous Carbon 2 behavior?

```suggestion
                // cast to int retain Carbon 2 behaviour of using whole days
                // pass `$absolute = true` argument to retain Carbon 2 behaviour of returning an absolute diff
```

tests/Jobs/PlatformStatsSummaryJobTest.php
```diff
@@ -274,4 +274,50 @@ public function testCreationStats() {
...
+            ], 200),
+            $this->mwBackendHost . '/w/api.php?action=query&list=allpages&apnamespace=120&apcontinue=&aplimit=max&format=json' => Http::response([
+                'query' => ['allpages' => []],
+            ], 200),
+        ]);
+
+        $job = new PlatformStatsSummaryJob;
+        (function ($resolver): void {
+            $this->mwHostResolver = $resolver;
+        })->call($job, $this->mockMwHostResolver);
```
outdooracorn, line 308: This is some gnarly code 😅  Am I right in thinking that it runs `PlatformStatsSummaryJob` with the private `$mwHostResolver` property replaced with a mock?

Did you copy this pattern from anywhere else? 

It looks like other parts of this test use something like this:
```
        $job = new PlatformStatsSummaryJob();
        $job->handle($manager, $this->mockMwHostResolver);
```

I find this a lot easier to read, is there a good reason not to do the same here?

tests/Jobs/SendEmptyWikiNotificationsJobTest.php
```diff
@@ -50,6 +50,24 @@ public function testEmptyWikiNotificationsSendNotification() {
         );
     }
 
+    // empty wikis, that are almost old enough (29 days and 23 hrs)
+    public function testEmptyWikiNotificationsNotSendNotification() {
```
outdooracorn: Optional nitpick


```suggestion
    public function testEmptyWikiNotificationsNotificationNotSent() {
```

tests/Jobs/SendEmptyWikiNotificationsJobTest.php
```diff
@@ -38,7 +38,7 @@ public function testEmptyWikiNotificationsSendNotification() {
         Notification::fake();
         $user = User::factory()->create(['verified' => true]);
         $wiki = Wiki::factory()->create(['created_at' => $thresholdDaysAgo]);
-        $manager = WikiManager::factory()->create(['wiki_id' => $wiki->id, 'user_id' => $user->id]);
+        WikiManager::factory()->create(['wiki_id' => $wiki->id, 'user_id' => $user->id]);
```
outdooracorn, line 41: This is a good clean-up change!

However, it is unrelated to the main purpose of this PR (improving code and tests after Carbon 2 -> 3 migration). In the future, I'd prefer it if changes like this (code clean-up, refactoring, or other improvements that aren't directly tied to the PR's primary objective) are submitted in a separate pull request.

Rationale:
* It keeps PRs focused and easier to review. E.g. the reviewer doesn't have to:
  * Spend time figuring out if this change is related to the task/PR or not.
  * Review if the unrelated change will have an unwanted side effect. This increases the scope of the review which slows the review process down.
* It makes it simpler to track the history of specific changes (e.g. why this variable was removed).
* If a regression occurs, it's easier to identify which change caused it and only the problematic changes need to be reverted reducing the likelihood that beneficial changes are also lost.
* It allows focused discussions about each change without mixing concerns.

I won't block the PR this time, please consider this for the future.

Thanks

tests/Jobs/SendEmptyWikiNotificationsJobTest.php
```diff
@@ -50,6 +50,24 @@ public function testEmptyWikiNotificationsSendNotification() {
...
+
+        Notification::fake();
+        $user = User::factory()->create(['verified' => true]);
+        $wiki = Wiki::factory()->create(['created_at' => $thresholdDaysAgo]);
+        WikiManager::factory()->create(['wiki_id' => $wiki->id, 'user_id' => $user->id]);
+        $wiki->wikiLifecycleEvents()->updateOrCreate(['first_edited' => null]);
+
+        $job = new SendEmptyWikiNotificationsJob;
+        $this->assertFalse($job->checkIfWikiIsOldAndEmpty($wiki));
+        $job->handle();
```
outdooracorn, line 67: Testing implementation details (like internal methods) through direct method calls is an antipattern that creates brittle, tightly-coupled tests.
* **Tests should verify behavior, not implementation**: The public API of `SendEmptyWikiNotificationsJob` is (or should be) the `handle()` method. Testing `checkIfWikiIsOldAndEmpty()` directly means the test is checking *how* the job works internally, not *what* it does.
* **Tight coupling to the implementation**: If the logic is later refactored (e.g. if `checkIfWikiIsOldAndEmpty()` changes its signature or is extracted to a reusable component), the test breaks even though the behavior hasn't changed.

Given that you have already called `Notification::fake()`, following the same pattern as the existing tests of asserting that the correct notifications have been sent would make for a better test. Also see https://laravel.com/docs/11.x/notifications#testing.

```suggestion
        $job = new SendEmptyWikiNotificationsJob;
        $job->handle();

        Notification::assertNothingSent();
```

---

Side-note (**NOT FOR THIS PR**): Quickly looking at `SendEmptyWikiNotificationsJob` I think only the `handle()` method should be `public` and the rest made `private` to avoid the implementation details leaking into the public interface.

outdooracorn (changes requested):

tests/Routes/Wiki/ConversionMetricTest.php
dati18: I did it thinking it's related to false-to-null change in Carbon 3 but clearly it's not a parse or factory method. I will reverse these changes

tests/Jobs/PlatformStatsSummaryJobTest.php
dati18: Yes this pattern does exactly what you said. The block of code binds `$this` to the `PlatformStatsSummaryJob` instance `$job` and sets $mwHostResolver property to my mock. In line 305, `$job->prepareStats()` uses the that injected resolver.

Why not `$job->handle($manager, $this->mockMwHostResolver);`? Because `handle()` is too much for this test setup, it runs the full job flow (DB connections, query building, etc) for `PlatformStatsSummaryJob` and expects the database to be ready to exist already. I bumped into "`prefix_site_stats` table missing" error trying the old approach.

```shell
PDOException: SQLSTATE[42S02]: Base table or view not found: 1146 Table 'mwdb_threshold_1.prefix_site_stats' doesn't exist
```

Useful reference:
- `Closure::call`
https://www.php.net/manual/en/closure.call.php
This is exactly what the gnarly code uses: temporarily bind `$this` and class scope, then execute the closure.

other tests in this class also marked themselves as "skipped"
```php
<?php
$this->markTestSkipped('Pollutes the deleted wiki list');
```
It seems like they interfered with the rest of the test suite in some way, especially the deleted-wiki list

dati18: If you want a cleaner version, I will have to write something like `injectMockMwHostResolver(PlatformStatsSummaryJob $job)` as a workaround and then call:
```php
         $job = new PlatformStatsSummaryJob;
         $this->injectMockMwHostResolver($job);
```
basically the same

outdooracorn: Ahh so it doesn't call the job like I said in my previous comment, it only calls the closure that replaces the private property.

> Why not `$job->handle($manager, $this->mockMwHostResolver);`? Because `handle()` is too much for this test setup, it runs the full job flow (DB connections, query building, etc) for `PlatformStatsSummaryJob` and expects the database to be ready to exist already. I bumped into "`prefix_site_stats` table missing" error trying the old approach.

That makes me think we aren't stubbing/mocking the dependencies well enough in the test and/or the Job needs refactoring.

> other tests in this class also marked themselves as "skipped"

Ahh good spot, I hadn't noticed that. :+1:  

> If you want a cleaner version, I will have to write something like `injectMockMwHostResolver(PlatformStatsSummaryJob $job)` as a workaround and then call:

Nah, that just moves the code to a different place.

---

Given that none of the other tests aren't actually run :see_no_evil: I'd be happy to settle for a comment in the code here. How about something like this?

```suggestion

        // This is a hack to override the `private` `PlatformStatsSummaryJob::mwHostResolver` property.
        // See https://www.php.net/manual/en/closure.call.php for more details on how this works.
        // TODO: figure out how to stub the `DatabaseManager` correctly and/or refactor the Job so that 
        // we can more easily inject dependencies in the tests.
        (function ($resolver): void {
            $this->mwHostResolver = $resolver;
        })->call($job, $this->mockMwHostResolver);
```

dati18: Agree to add the comment; it looks less ugly and less scary with a bit of comments explaining.

dati18: `lint` is not happy with blank space. So remember to check for it next time :)

```
  -        // TODO: figure out how to stub the `DatabaseManager` correctly and/or refactor the Job so that 
  +        // TODO: figure out how to stub the `DatabaseManager` correctly and/or refactor the Job so that
```

outdooracorn: > `lint` is not happy with blank space. So remember to check for it next time :)

Oh, that's an interesting one! We should probably have a team discussion around who is responsible for the correctness in a PR, the author or a reviewer that left a suggestion.

I would say that as the author you are responsible, and it's up to you if you take the _suggestion_ as is and rely on CI to do tests and linting; OR you apply the suggestion locally, and run tests and linting locally before pushing to GitHub.

outdooracorn (approved):

PR #1153, opened 2026-06-03
Fix data providers in ProfileValidatorTest

outdooracorn:
The data providers in `ProfileValidatorTest` where returning an improperly constructed array causing the data set name to b treated as a named parameter (which obviously didn't exist). This was tolerated but deprecated in PHPUnit 10.x. It will not be supported in PHPUnit 11.x.

Return generator from data providers for less array nesting and easier reading.

Bug: T428034

### DIFF
--- a/tests/Helper/ProfileValidatorTest.php
+++ b/tests/Helper/ProfileValidatorTest.php
@@ -24,46 +24,46 @@ class ProfileValidatorTest extends TestCase {
         $this->assertFalse($validator->passes());
     }
 
-    public static function validProfileProvider() {
-        return [
-            ['boring profile with no other' => [
-                'purpose' => 'data_hub',
-                'audience' => 'narrow',
-                'temporality' => 'permanent',
-            ]],
-            ['with other values' => [
-                'purpose' => 'data_hub',
-                'audience' => 'other',
-                'audience_other' => 'my cat',
-                'temporality' => 'other',
-                'temporality_other' => 'only in the past',
-            ]],
-        ];
+    public static function validProfileProvider(): iterable {
+        yield 'boring profile with no other' => [[
+            'purpose' => 'data_hub',
+            'audience' => 'narrow',
+            'temporality' => 'permanent',
+        ]];
+
+        yield 'with other values' => [[
+            'purpose' => 'data_hub',
+            'audience' => 'other',
+            'audience_other' => 'my cat',
+            'temporality' => 'other',
+            'temporality_other' => 'only in the past',
+        ]];
     }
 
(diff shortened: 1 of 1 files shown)


tests/Helper/ProfileValidatorTest.php
dati18, line 1: For `ProfileValidatorTest`, `array` is perfectly fine and probably the most readable, simple and familiar.
Using `Generator` here is optional but I don't think it's necessary because our test dataset is small
The important part for this deprecation fix is dataset shape
I suggest this instead

```php
    public static function validProfileProvider() {
        return [
            'boring profile with no other' => [[
                'purpose' => 'data_hub',
                'audience' => 'narrow',
                'temporality' => 'permanent',
            ]],
            'with other values' => [[
                'purpose' => 'data_hub',
                'audience' => 'other',
                'audience_other' => 'my cat',
                'temporality' => 'other',
                'temporality_other' => 'only in the past',
            ]],
        ];
    }

    public static function invalidProfileProvider() {
        return [
            'missing other keys' => [[
                'purpose' => 'data_hub',
                'audience' => 'narrow',
                'temporality' => 'other',
            ]],
            'audience is empty string' => [[
                'purpose' => 'data_hub',
                'audience' => '',
                'temporality' => 'other',
            ]],
            'audience key present when purpose not data_hub' => [[
                'purpose' => 'data_lab',
                'audience' => 'narrow',
                'temporality' => 'permanent',
            ]],
            'other keys when there should not be' => [[
                'purpose' => 'data_hub',
                'purpose_other' => 'asdfasdf',
                'audience' => 'narrow',
                'temporality' => 'permanent',
            ]],
        ];
    }

```

dati18 (changes requested):

tests/Helper/ProfileValidatorTest.php
outdooracorn: I'd like us to move to using generators over arrays as I see them having several benefits:
* Less nesting required which means you can fit more code on a single line, less need to find nice ways to wrap code onto multiple lines, easier to read.
* It's easier to group related variables and `yield` statement all together for easier reading whereas `return [...]` must go at the end of the data provider method (granted you can get around that with `$dataset = []; $dataset['description'] = ...; return $dataset;` but that is messier and more verbose).
* Easier to programmatically build and compose data sets.
* Faster startup and lower memory usage for large data sets.

PHPUnit has supported generator data providers for a while now. I encourage you to become familiar with them if you are not already.

I don't believe this suggestion is worth the effort, so I won't change the PR.

dati18 (approved):

PR #1156, opened 2026-06-10
Update PublicWikiController parameter validation and add unit tests

outdooracorn:
* Validate `page` and `per_page` params as integers rather than numbers
* Add `page` and `per_page` minimum limit of 1
* Add `per_page` maximum limit of 100

Bug: T428711

### DIFF
--- a/app/Http/Controllers/PublicWikiController.php
+++ b/app/Http/Controllers/PublicWikiController.php
@@ -24,8 +24,8 @@ class PublicWikiController extends Controller {
             'direction' => 'in:desc,asc',
             'is_featured' => 'boolean',
             'is_active' => 'boolean',
-            'per_page' => 'numeric',
-            'page' => 'numeric',
+            'per_page' => 'int|min:1|max:100',
+            'page' => 'int|min:1',
         ]);
 
         $params = array_merge(self::$defaultParams, $request->input());
--- a/app/Http/Controllers/PublicWikiController.php
+++ b/app/Http/Controllers/PublicWikiController.php
@@ -11,6 +11,7 @@ class PublicWikiController extends Controller {
     private static $defaultParams = [
         'sort' => 'sitename',
         'direction' => 'asc',
+        'per_page' => 15,
     ];
 
     private static $activeThresholdPageCount = 2;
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/PublicWikiController.php
```diff
@@ -24,8 +24,8 @@ public function index(Request $request) {
             'direction' => 'in:desc,asc',
             'is_featured' => 'boolean',
             'is_active' => 'boolean',
-            'per_page' => 'numeric',
-            'page' => 'numeric',
+            'per_page' => 'int|min:1|max:100',
```
outdooracorn, line 28: Discuss and confirm `100` is the max value we want to use

outdooracorn (commented):
WIP until we have informed the suite team ~~and they have changed https://github.com/wmde/wikibase-metadata/blob/748652566556d1255e8dc79c21b64bd58c2b6bb9/backend/fetch_data/cloud_api_data/get_cloud_api_data.py#L7~~

app/Http/Controllers/PublicWikiController.php
tarrow: 100 is totally fine by me

app/Http/Controllers/PublicWikiController.php
tarrow: I wonder if we can now remove this

outdooracorn: If we add a default value for `per_page` we can. I chose 15 as that seems to be what the default is if `null` is used.

tarrow (dismissed):

tests/Http/Controllers/PublicWikiControllerTest.php
```diff
@@ -35,6 +38,44 @@ public function testShowLoadsWikiLatestProfileForResource(): void {
...
+        yield 'page is not int' => [['page' => 2.3], true];
+        yield 'page is too low' => [['page' => 0], true];
+        yield 'page is min value' => [['page' => 1], false];
+    }
+
+    /**
+     * @dataProvider provideQueryParamsAndErrorExpected
+     */
+    public function testIndexQueryParamValidation(array $queryParams, bool $errorExpected): void {
+        $controller = new PublicWikiController;
```
tarrow:
```suggestion
        $controller = new PublicWikiController();
```

tarrow (approved):
boldly approving my own lint change

PR #1157, opened 2026-06-10
lint: require parentheses when instantiating a new class

outdooracorn:
Set the `laravel/pint` config `new_with_parentheses` to `true` to make it more explicit that a new instance is created.

Bug: T428714

### DIFF
--- a/app/Console/Commands/Invitation/CreateBulk.php
+++ b/app/Console/Commands/Invitation/CreateBulk.php
@@ -18,7 +18,7 @@ class CreateBulk extends Command {
      */
     public function handle() {
         $numCodes = trim($this->argument('numCodes'));
-        $helper = new InviteHelper;
+        $helper = new InviteHelper();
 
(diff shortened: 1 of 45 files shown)


outdooracorn (commented):
I personally find this more explicit that a new instance is created. I've had times when I've missed that a class was instantiated (vs just a reference to the class). This is likely because I come from Python where the parenthesis are used to differentiate between instantiation of a class and just a reference to a class:

```
class MyClass:
    def __init__(self):
      print("MyClass was intantiated")

a = MyClass()  # instance
b = MyClass  # reference

print(a)
print(b)

## Outputs
# MyClass was intantiated
# <__main__.MyClass object at 0x7d5a4ec09bb0>
# <class '__main__.MyClass'>

tarrow (approved):
I have no objections to this style change. I think it is mostly personal preference and I don't have a strong opinion

PR #1158, opened 2026-06-10
tests: prevent output in phpunit tests

outdooracorn:
I noticed an accidental `echo` left in a test during a recent review. It's best practice to avoid output during tests. Update PHPUnit config to make CI fail, rather than relying on human code review.

* Modify `ElasticSearchIndexDeleteTest` so it doesn't print output.
* `beStrictAboutOutputDuringTests` marks tests as "risky" if there is output printed.
* `failOnRisky` makes the test suite fail if any tests are marked as "risky".

Bug: T428716

### DIFF
--- a/tests/Jobs/Integration/ElasticSearchIndexDeleteTest.php
+++ b/tests/Jobs/Integration/ElasticSearchIndexDeleteTest.php
@@ -48,12 +48,11 @@ class ElasticSearchIndexDeleteTest extends TestCase {
         );
         $response = $curlRequest->execute();
         $err = $curlRequest->error();
-        var_dump($response);
+        $curlRequest->close();
 
         if ($err) {
-            var_dump($err);
+            $this->fail("Request returned an error: {$err}");
         }
-        $curlRequest->close();
 
         return json_decode($response, true);
     }
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -9,6 +9,8 @@
   xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/10.5/phpunit.xsd"
   cacheDirectory=".phpunit.cache"
   backupStaticProperties="false"
+  beStrictAboutOutputDuringTests="true"
+  failOnRisky="true"
 >
   <testsuites>
     <testsuite name="Application Test Suite">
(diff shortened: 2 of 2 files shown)


dati18 (dismissed):
approved.

tests/Jobs/Integration/ElasticSearchIndexDeleteTest.php
```diff
@@ -48,12 +48,11 @@ public function makeRequest($url, $method = 'GET') {
         );
         $response = $curlRequest->execute();
         $err = $curlRequest->error();
-        var_dump($response);
+        $curlRequest->close();
```
tarrow, line 51: I'm not sure I understand the purpose of this. From the little searching I could see since PHP 8.0 we shouldn't close resources (since this isn't even a resource) any more. Sadly my only sources were blogs like [1](https://phpbackend.com/blog/post/php-8-0-curlHandle-object-in-curl-functions) and [2](https://php.watch/versions/8.0/resource-CurlHandle) and the rather sparse docs on the [topic](https://www.php.net/manual/en/function.curl-close.php).

Do we need to do this? I also note that there is quite some layer of indirection in calling this via our own wrapper class

tests/Jobs/Integration/ElasticSearchIndexDeleteTest.php
```diff
@@ -48,12 +48,11 @@ public function makeRequest($url, $method = 'GET') {
...
 
         if ($err) {
-            var_dump($err);
+            $this->fail("Request returned an error: {$err}");
```
tarrow, line 54: I was confused for a while why the test doesn't fail here given that it *expects* to get a 404 at some points. It doesn't error because we haven't set the option for that behaviour: https://curl.se/libcurl/c/CURLOPT_FAILONERROR.html

tarrow (changes requested):
Would you mind just clarifying the `->close()`?

tests/Jobs/Integration/ElasticSearchIndexDeleteTest.php
tarrow: We *don't* need this since php 8.

What I also overlooked was that this is not new code but moved up from below the fold of the page.

Let's move ahead with this and we can refactor this test later if needed

tarrow (dismissed):

tarrow (approved):
approve change of base

PR #1162, opened 2026-06-15
default to JSON responses for exceptions

tarrow:
Overwrite the default error handler in our custom handler to so it shouldReturnJson is always true.

This prevents http responses (and the confusing redirect) with ValidationExceptions

Bug: T429021

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -25,4 +25,8 @@ class Handler extends ExceptionHandler {
             (new ErrorReporting())->report($e);
         });
     }
+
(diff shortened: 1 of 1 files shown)


deer-wmde (commented):
nice find! I love how clean this is

Can confirm it works locally:
```
$ curl -s --insecure "https://www.wbaas.dev/api/wiki?per_page=101" 
{"message":"The per page must not be greater than 100.","errors":{"per_page":["The per page must not be greater than 100."]}}
```

deer-wmde (approved):

PR #1168, opened 2026-06-18
Resolve GuzzleHttp\Exception\RequestException error

dati18:
[bumping guzzlehttp/guzzle from 7.10.5 to 7.12.0 ]( https://github.com/wbstack/api/pull/1165) triggered a bunch of [error](https://console.cloud.google.com/errors/detail/CNrO9YzO99qlkQE?project=wikibase-cloud&supportedpurview=project)

The stack trace pointed to line 21 and 30 of [PollForMediaWikiJob](https://github.com/wbstack/api/blob/main/app/Jobs/PollForMediaWikiJobsJob.php#L30)

The investigation is documented [here](https://phabricator.wikimedia.org/T429593#12058620).

Bug: T429593

### DIFF
--- a/app/Metrics/App/WikiMetrics.php
+++ b/app/Metrics/App/WikiMetrics.php
@@ -79,7 +79,7 @@ class WikiMetrics {
             return null;
         }
 
-        $endpoint = 'http://' . trim($qsNamespace->backend, '/')
+        $endpoint = 'http://' . $qsNamespace->backend
             . '/bigdata/namespace/'
             . $qsNamespace->namespace
             . '/sparql';
--- a/app/Metrics/App/WikiMetrics.php
+++ b/app/Metrics/App/WikiMetrics.php
@@ -106,7 +106,7 @@ class WikiMetrics {
     protected function normalizeBackendUrl(string $backend): string {
         $backend = rtrim($backend, '/');
 
-        if (str_starts_with($backend, 'http://') || str_starts_with($backend, 'https://')) {
+        if (filter_var($backend, FILTER_VALIDATE_URL) !== false) {
             return $backend;
         }
 
--- a/app/Metrics/App/WikiMetrics.php
+++ b/app/Metrics/App/WikiMetrics.php
@@ -100,6 +100,9 @@ class WikiMetrics {
         return null;
     }
 
+    /**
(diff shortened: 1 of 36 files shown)


tarrow (dismissed):
https://github.com/wbstack/api/pull/1168#issuecomment-4831649874

tarrow (changes requested):
Thanks; great work. This is indeed very complex https://github.com/wbstack/api/pull/1168#issuecomment-4856354663

tests/Services/MediaWikiHostResolverTest.php
```diff
@@ -53,4 +64,13 @@ public function testResolverThrowsIfUnableToFindWiki(): void {
...
     }
+
+    public function testLaravelHttpAcceptsSchemeUrls(): void {
+        Http::fake(['*' => Http::response()]);
+
+        // Using Laravel Http facade (which properly respects Http::fake())
+        $response = Http::get('http://mediawiki-143-app-backend.default.svc.cluster.local/w/api.php');
+
+        $this->assertTrue($response->status() === 200);
+    }
```
tarrow:
```suggestion

```

I don't think this test belongs here (or anywhere) because I believe it's entirely testing the behaviour of Laravel's `Http::get()` and none of our code.

app/Metrics/App/WikiMetrics.php
```diff
@@ -79,7 +79,10 @@ protected function getNumOfTriples(): ?int {
             return null;
         }
 
-        $endpoint = $qsNamespace->backend . '/bigdata/namespace/' . $qsNamespace->namespace . '/sparql';
+        $endpoint = 'http://' . trim($qsNamespace->backend, '/')
```
outdooracorn: How come this `trim()`? I don't think `$qsNamespace->backend` ever ends in a `/`.

dati18: just extra defensive normalization. A host value coming from `config('app.queryservice_host')` shouldn't have a `/` at the end. Thanks for pointing it out :D

tests/Services/MediaWikiHostResolverTest.php
dati18: leftover test case, deleted

tarrow (approved):
looks ok to me now; let's ship to staging and keep a sharp eye out for any new errors

PR #1172, opened 2026-06-22
chore: fix warnings from docker build step

tarrow:
There are warnings in the docker build step. These are also shown in the PR review UI.

Resolving them means that we will have less noise to read when reviewing PRs and it means we're better following docker best practices.

Bug: T429818


dati18 (approved):
I noticed the noise also. This is a welcoming change. Approved

PR #1177, opened 2026-06-24
Create schemas and models for implementing generic policies

tarrow:
Models, migrations, and simple tests for Policy and PolicyAcceptance. These will be used to track which policies exist over the lifetime of the platform and when each user accepted them.

Use built in `created_at` and `updated_at` timestamps via the `$table->timestamps()` method, for consistency with our other models.

Does not include Factories or Seeders. Possibility for a follow-up.

Bug: T428175

### DIFF
--- a/app/Policy.php
+++ b/app/Policy.php
@@ -45,7 +45,6 @@ class Policy extends Model {
             // cast `active_from` to a CarbonImmutable instance rather than a string
             'active_from' => 'immutable_date',
 
-            // TODO: should we make Laravel use CarbonImmutable globally instead of casting in models?
             'created_at' => 'immutable_datetime',
             'updated_at' => 'immutable_datetime',
         ];
--- a/app/Policy.php
+++ b/app/Policy.php
@@ -8,12 +8,12 @@ use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Model;
 
 /**
- * @property-read int $id
+ * @property int $id
  * @property string $policy_type
  * @property CarbonImmutable|null $active_from
  * @property string $content_vue_file
- * @property CarbonImmutable $created_at
- * @property CarbonImmutable $updated_at
+ * @property CarbonImmutable|null $created_at
+ * @property CarbonImmutable|null $updated_at
  *
  * @method static Builder<static>|Policy newModelQuery()
  * @method static Builder<static>|Policy newQuery()
--- a/app/Policy.php
+++ b/app/Policy.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace App;
+
+use Carbon\CarbonImmutable;
+use Eloquent;
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Model;
+
+/**
+ * @property-read int $id
+ * @property string $policy_type
+ * @property CarbonImmutable|null $active_from
+ * @property string $content_vue_file
+ * @property CarbonImmutable $created_at
+ * @property CarbonImmutable $updated_at
+ *
+ * @method static Builder<static>|Policy newModelQuery()
+ * @method static Builder<static>|Policy newQuery()
+ * @method static Builder<static>|Policy query()
+ * @method static Builder<static>|Policy whereActiveFrom($value)
+ * @method static Builder<static>|Policy whereContentVueFile($value)
+ * @method static Builder<static>|Policy whereCreatedAt($value)
+ * @method static Builder<static>|Policy whereId($value)
+ * @method static Builder<static>|Policy wherePolicyType($value)
+ * @method static Builder<static>|Policy whereUpdatedAt($value)
+ *
+ * @mixin Eloquent
+ */
+class Policy extends Model {
+    // TODO: also create a factory (and seeder)?
+
+    // define which attributes are mass assignable
+    protected $fillable = [
+        'policy_type',
+        'active_from',
+        'content_vue_file',
+    ];
+
+    // define the default value of model attributes when a new instance is created
+    protected $attributes = [
+        'active_from' => null,
+    ];
+
+    protected function casts(): array {
+        return [
+            // cast `active_from` to a CarbonImmutable instance rather than a string
+            'active_from' => 'immutable_date',
+
+            // TODO: should we make Laravel use CarbonImmutable globally instead of casting in models?
+            'created_at' => 'immutable_datetime',
+            'updated_at' => 'immutable_datetime',
+        ];
+    }
+}
--- a/app/Policy.php
+++ b/app/Policy.php
@@ -42,9 +42,10 @@ class Policy extends Model {
 
     protected function casts(): array {
         return [
-            // cast `active_from` to a CarbonImmutable instance rather than a string
+            // cast `active_from` to a `CarbonImmutable` instance rather than a string
             'active_from' => 'immutable_date',
 
+            // cast to `CarbonImmutable` until we default to using `CarbonImmutable` globally in T430656
             'created_at' => 'immutable_datetime',
             'updated_at' => 'immutable_datetime',
         ];
--- a/app/Policy.php
+++ b/app/Policy.php
@@ -28,8 +28,6 @@ use Illuminate\Database\Eloquent\Model;
  * @mixin Eloquent
  */
 class Policy extends Model {
-    // TODO: also create a factory (and seeder)?
-
     // define which attributes are mass assignable
     protected $fillable = [
         'policy_type',
--- a/app/PolicyAcceptance.php
+++ b/app/PolicyAcceptance.php
@@ -15,9 +15,9 @@ use Illuminate\Database\Eloquent\Model;
  * @property int $id
  * @property int $user_id
  * @property int $policy_id
+ * @property CarbonImmutable $accepted_at
  * @property CarbonImmutable|null $created_at
  * @property CarbonImmutable|null $updated_at
- * @property CarbonImmutable $accepted_at
  *
  * @method static Builder<static>|PolicyAcceptance newModelQuery()
  * @method static Builder<static>|PolicyAcceptance newQuery()
--- a/app/PolicyAcceptance.php
+++ b/app/PolicyAcceptance.php
@@ -9,7 +9,7 @@ use Illuminate\Database\Eloquent\Model;
 
 /**
  * This model uses a separate `accepted_at` property rather than renaming the default `created_at` property because:
- *   - it reduces confuses by remaining consistent with other models that use the default timestamps
+ *   - it remains consistent with other models that use the default timestamps
  *   - `accepted_at` will be before `created_at` when backfilling the terms-of-use acceptances
  *
  * @property int $id
@@ -35,7 +35,6 @@ class PolicyAcceptance extends Model {
     protected $fillable = [
         'user_id',
         'policy_id',
-
     ];
 
     protected $guarded = [
--- a/app/PolicyAcceptance.php
+++ b/app/PolicyAcceptance.php
@@ -32,14 +32,16 @@ use Illuminate\Database\Eloquent\Model;
  * @mixin Eloquent
  */
 class PolicyAcceptance extends Model {
-    // TODO: also create a factory (and seeder)?
-
-    // define which attributes are mass assignable
     protected $fillable = [
         'user_id',
         'policy_id',
-        // Don't allow `accepted_at` to be mass assigned? Most of the time this will be set to the current timestamp by the database.
-        // 'accepted_at',
+
+    ];
+
+    protected $guarded = [
+        // Don't allow `accepted_at` to be mass assigned.
+        // Most of the time this will be set to the current timestamp by the database.
+        'accepted_at',
     ];
 
     protected function casts(): array {
@@ -48,6 +50,7 @@ class PolicyAcceptance extends Model {
             'accepted_at' => 'immutable_datetime',
 
             // TODO: should we make Laravel use CarbonImmutable globally instead of casting in models?
+            // TODO: should we do any casting with these built ins?
             'created_at' => 'immutable_datetime',
             'updated_at' => 'immutable_datetime',
         ];
--- a/app/PolicyAcceptance.php
+++ b/app/PolicyAcceptance.php
@@ -12,12 +12,12 @@ use Illuminate\Database\Eloquent\Model;
  *   - it reduces confuses by remaining consistent with other models that use the default timestamps
  *   - `accepted_at` will be before `created_at` when backfilling the terms-of-use acceptances
  *
- * @property-read int $id
+ * @property int $id
  * @property int $user_id
  * @property int $policy_id
+ * @property CarbonImmutable|null $created_at
+ * @property CarbonImmutable|null $updated_at
  * @property CarbonImmutable $accepted_at
- * @property CarbonImmutable $created_at
- * @property CarbonImmutable $updated_at
  *
  * @method static Builder<static>|PolicyAcceptance newModelQuery()
  * @method static Builder<static>|PolicyAcceptance newQuery()
--- a/app/PolicyAcceptance.php
+++ b/app/PolicyAcceptance.php
@@ -35,11 +35,6 @@ class PolicyAcceptance extends Model {
     protected $fillable = [
         'user_id',
         'policy_id',
-    ];
-
-    protected $guarded = [
-        // Don't allow `accepted_at` to be mass assigned.
-        // Most of the time this will be set to the current timestamp by the database.
         'accepted_at',
     ];
 
@@ -48,6 +43,7 @@ class PolicyAcceptance extends Model {
             // cast `accepted_at` to a `CarbonImmutable` instance rather than a string
             'accepted_at' => 'immutable_datetime',
 
+            // cast to `CarbonImmutable` until we default to using `CarbonImmutable` globally in T430656
             'created_at' => 'immutable_datetime',
             'updated_at' => 'immutable_datetime',
         ];
--- a/app/PolicyAcceptance.php
+++ b/app/PolicyAcceptance.php
@@ -49,8 +49,6 @@ class PolicyAcceptance extends Model {
             // cast `accepted_at` to a `CarbonImmutable` instance rather than a string
             'accepted_at' => 'immutable_datetime',
 
-            // TODO: should we make Laravel use CarbonImmutable globally instead of casting in models?
-            // TODO: should we do any casting with these built ins?
             'created_at' => 'immutable_datetime',
             'updated_at' => 'immutable_datetime',
         ];
--- a/app/PolicyAcceptance.php
+++ b/app/PolicyAcceptance.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace App;
+
+use Carbon\CarbonImmutable;
+use Eloquent;
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Model;
+
+/**
+ * This model uses a separate `accepted_at` property rather than renaming the default `created_at` property because:
+ *   - it reduces confuses by remaining consistent with other models that use the default timestamps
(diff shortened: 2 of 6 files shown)


tests/PolicyAcceptanceTest.php
```diff
@@ -0,0 +1,49 @@
...
+use App\Policy;
+use App\PolicyAcceptance;
+use App\User;
+use Carbon\CarbonImmutable;
+use Illuminate\Foundation\Testing\RefreshDatabase;
+
+class PolicyAcceptanceTest extends TestCase {
+    use RefreshDatabase;
+
+    protected int $user_id;
```
dati18: property name with `snake_case` is a bit odd, init? I think it should be `$userId`  or `$policyId` for consistency and for the sake of good php practice

tests/PolicyTest.php
```diff
@@ -0,0 +1,29 @@
...
+
+use App\Policy;
+use Carbon\Carbon;
+use Illuminate\Foundation\Testing\RefreshDatabase;
+
+class PolicyTest extends TestCase {
+    use RefreshDatabase;
+
+    public function testCreatesAndSavesSuccessfully(): void {
+        $yesterday = Carbon::yesterday();
```
dati18: nitpick: `yesterday()` is a bit time-dependent, I suggest using a fixed point in time so the test is deterministic.

tests/PolicyTest.php
```diff
@@ -0,0 +1,29 @@
...
+        $policy = Policy::create(
+            [
+                'policy_type' => 'terms-of-use',
+                'active_from' => $yesterday,
+                'content_vue_file' => 'terms-of-use/example.vue',
+            ]
+        );
+        $policy->save();
```
dati18: `create()` already saves, so is this line redundant?

dati18: If you remove this line, then the test case should be renamed too.

app/Policy.php
```diff
@@ -0,0 +1,55 @@
...
+    // define which attributes are mass assignable
+    protected $fillable = [
+        'policy_type',
+        'active_from',
+        'content_vue_file',
+    ];
+
+    // define the default value of model attributes when a new instance is created
+    protected $attributes = [
+        'active_from' => null,
```
dati18, line 40: Do unset Eloquent attributes default to `null` anyway?

app/PolicyAcceptance.php
```diff
@@ -0,0 +1,55 @@
...
+namespace App;
+
+use Carbon\CarbonImmutable;
+use Eloquent;
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Model;
+
+/**
+ * This model uses a separate `accepted_at` property rather than renaming the default `created_at` property because:
+ *   - it reduces confuses by remaining consistent with other models that use the default timestamps
```
dati18: do you mean "reduces confusion"?

dati18 (changes requested):

app/PolicyAcceptance.php
```diff
@@ -0,0 +1,56 @@
...
+ * @mixin Eloquent
+ */
+class PolicyAcceptance extends Model {
+    protected $fillable = [
+        'user_id',
+        'policy_id',
+
+    ];
+
+    protected $guarded = [
```
dati18: `accepted_at` is already not mass-assignable because it isn’t listed in `$fillable`, right?

app/PolicyAcceptance.php
```diff
@@ -0,0 +1,56 @@
...
+ * @method static Builder<static>|PolicyAcceptance whereUpdatedAt($value)
+ * @method static Builder<static>|PolicyAcceptance whereUserId($value)
+ *
+ * @mixin Eloquent
+ */
+class PolicyAcceptance extends Model {
+    protected $fillable = [
+        'user_id',
+        'policy_id',
+
```
dati18: remove think blank line

dati18 (changes requested):

tests/PolicyTest.php
dati18: then you also don't need to assign `Policy::create()` to `$policy` and just call it

tests/PolicyAcceptanceTest.php
```diff
@@ -0,0 +1,49 @@
...
+        );
+        $policyAcceptance->save();
+        $policyAcceptance->refresh();
+
+        $this->assertDatabaseHas('policy_acceptances', [
+            'user_id' => $this->user_id,
+            'policy_id' => $this->policy_id,
+        ]);
+
+        $this->assertNotEmpty($policyAcceptance->accepted_at);
```
dati18, line 45: I can think of one more assertion to assert that `policyAcceptance` is an instance of `CarbonImmuatble`

```
$this->assertInstanceOf(CarbonImmutable::class, $policyAcceptance->accepted_at);
```

app/PolicyAcceptance.php
dati18: I can think of adding a test case (for example `testAcceptedAtIgnoresMassAssignment`) to confirm that `accepted_at` isn't mass-assigned

tests/PolicyAcceptanceTest.php
tarrow: yes! You're right. Fixed and we should add some linting check for it I think

tests/PolicyTest.php
tarrow: fair enough! Fixed to sometime last year

tests/PolicyTest.php
tarrow: good idea, getting myself muddled with `new` vs `::create()`

app/Policy.php
tarrow: yes they do, I liked the clarity of specifying it though

app/PolicyAcceptance.php
tarrow: I do, I also rephrased it a little more

app/PolicyAcceptance.php
tarrow: That's true. I also thought in this case it was nice to be explict that it isn't mass assignable.

I did actually debate in my head for a while if we really care to make this mass assignable or not. I'm not sure we buy much protection this way

In the end I continued with @outdooracorn 's decision to keep it that way but tried to make it clearer

app/PolicyAcceptance.php
tarrow: definitely possible and I have done so. If we want this style we should add a lint setting for it though

tests/PolicyAcceptanceTest.php
tarrow: good idea, added!

app/Policy.php
dati18: You're the boss, if you said so then let's keep it :)

app/PolicyAcceptance.php
tarrow: Didn't see the test comment, I have also added a test

tests/PolicyAcceptanceTest.php
```diff
@@ -0,0 +1,61 @@
...
+
+    public function testAcceptedAtIgnoresMassAssignment(): void {
+        $policyAcceptance = PolicyAcceptance::create(
+            [
+                'user_id' => $this->userId,
+                'policy_id' => $this->policyId,
+                'accepted_at' => CarbonImmutable::createFromDate(2026, 1, 1),
+            ]
+        );
+        $this->assertNull($policyAcceptance->accepted_at);
```
dati18: put the instance assertion here too is not a bad idea  `$this->assertInstanceOf(CarbonImmutable::class, $policyAcceptance->accepted_at);`

tarrow: I think that's not possible: `null` isn't an instance of `CarbonImmutable`

tests/PolicyAcceptanceTest.php
```diff
@@ -0,0 +1,61 @@
...
+        parent::setUp();
+        $user = User::factory()->create();
+        $this->userId = $user->id;
+        $policy = Policy::create(
+            [
+                'policy_type' => 'terms-of-use',
+                'active_from' => CarbonImmutable::yesterday(),
+                'content_vue_file' => 'terms-of-use/example.vue',
+            ]);
+        $policy->save();
```
dati18: `save()` here is not really necessary. I removed it and the test is still "green"

tarrow: cool! You're right. Also removed

database/migrations/2026_06_22_083853_create_policies_table.php
```diff
@@ -0,0 +1,32 @@
...
+     */
+    public function up(): void {
+        Schema::create('policies', function (Blueprint $table) {
+            $table->id();
+            $table->enum('policy_type', ['terms-of-use', 'hosting-policy']);
+            $table->date('active_from')->nullable()->default(null);
+            $table->string('content_vue_file', 255);
+
+            // Use Eloquent built in to create nullable `created_at` and `updated_at` timestamp fields
+            $table->timestamps();
```
outdooracorn, line 19: > @outdooracorn on your return can you take a look at [b40bc27](https://github.com/wbstack/api/commit/b40bc27378707982b58bb9afbf2fd7a22d3f56fb) and see if you agree with this change? I assume you very explicitly decided to try and make the built in timestamps (`created_at` and `updated_at`) non-nullable but I didn't understand why.
> 
> I did quite some research and from what I could read upstream making the built in timestamps nullable was a very explicit nearly decade old decision to handle mysql/maria.

Yes, sorry, I didn't have enough time to research this thoroughly or document why I specifically made them not-nullable.

The reason I explicitly created the `created_at` and `updated_at` columns as not-nullable was because I noticed that `$table->timestamps();` resulted in the following SQL:
```
`created_at` timestamp NULL DEFAULT NULL,
`updated_at` timestamp NULL DEFAULT NULL,
```
I found this odd as I would assume that we never want `created_at` or `updated_at` columns to be `NULL` and I would like to be notified early if some code (ours or Laravel's) was trying to set these values to `NULL`. Especially if the business logic relied on these timestamps being set.

---

The [issue linked](https://github.com/laravel/framework/issues/12490) in commit b40bc27 didn't seem to discuss the issue or rational that much. However, from my reading, the reason Laravel does `timestamp NULL DEFAULT NULL` is because of MySQL/MariaDB default behaviour. The whole [Automatic Values section of the MariaDB TIMESTAMP docs](https://mariadb.com/docs/server/reference/data-types/date-and-time-data-types/timestamp#automatic-values) is worth a read.

Another, useful resource is https://jira.mariadb.org/browse/MDEV-28632 which lists some examples of when this MariaDB behaviour was confusing.

I'm still not sure why Laravel chose to do
```
  `created_at` timestamp NULL DEFAULT NULL,
  `updated_at` timestamp NULL DEFAULT NULL,
```
rather than
```
  `created_at` timestamp NOT NULL DEFAULT current_timestamp(),
  `updated_at` timestamp NOT NULL DEFAULT current_timestamp() ON UPDATE current_timestamp(),
```
but I guess this is because Eloquent correctly handles setting the value of `created_at` and `updated_at` when `INSERT`ing or `UPDATE`ing data. So even though the columns are nullable and default to `NULL`, the data in the columns should (hopefully) never actually be `NULL`.

---

**In any case, I'm happy with using the default `$table->timestamps();` for now.**

---

While investigating, I noticed a few things in `database/schema/mysql-schema.sql` that didn't look right:
* The `tou_acceptance` table has
`'tou_accepted_at' timestamp NOT NULL DEFAULT current_timestamp() ON UPDATE current_timestamp(),` which could be an issue if existing data in that table was ever modified, and the table wasn't about to be deleted.
* The `password_resets` table has `'created_at' timestamp NOT NULL DEFAULT current_timestamp() ON UPDATE current_timestamp(),`. My initial assumption is that the `created_at` value should not be modified to use the `current_timestamp()` when a row is updated. I haven't dug into this at all, though.

database/migrations/2026_06_22_083910_create_policy_acceptances_table.php
```diff
@@ -0,0 +1,37 @@
...
+            $table->foreignId('policy_id')->constrained()->restrictOnUpdate()->restrictOnDelete();
+
+            // Use Eloquent built in to create nullable `created_at` and `updated_at` timestamp fields
+            $table->timestamps();
+
+            // Using a separate `accepted_at` column rather than renaming the default `created_at` column because:
+            //   * it reduces confusion by remaining consistent with other tables that use these default columns
+            //   * `accepted_at` will be before `created_at` when backfilling the terms-of-use acceptances
+            $table->timestamp('accepted_at')->useCurrent();
+        });
```
outdooracorn, line 33: Bit of a nitpick, but how come our explicitly created `accepted_at` column is after the Laravel default `timestamps()` columns?

tests/PolicyAcceptanceTest.php
```diff
@@ -0,0 +1,60 @@
...
+
+    public function testAcceptedAtIgnoresMassAssignment(): void {
+        $policyAcceptance = PolicyAcceptance::create(
+            [
+                'user_id' => $this->userId,
+                'policy_id' => $this->policyId,
+                'accepted_at' => CarbonImmutable::createFromDate(2026, 1, 1),
+            ]
+        );
+        $this->assertNull($policyAcceptance->accepted_at);
```
outdooracorn: FYI, this is only `null` until you read it from the database again e.g. with `policyAcceptance->refresh()` or `PolicyAcceptance::find($this->policyId)`. Not necessarily an issue with this test, but thought I would point out that behaviour as it was slightly confusing to me at first.

We could add something like this to help show that `accepted_at` is only temporarily `null`:

```suggestion
        $this->assertNull($policyAcceptance->accepted_at);
        $policyAcceptance->refresh();
        $this->assertNotEmpty($policyAcceptance->accepted_at);
        $this->assertInstanceOf(CarbonImmutable::class, $policyAcceptance->accepted_at);
```

but maybe that is overkill.

database/migrations/2026_06_22_083853_create_policies_table.php
```diff
@@ -0,0 +1,32 @@
...
+
+            // This prevents two upcoming policies of the same type with `active_from` set to `null`,
+            $table->unique(['policy_type', 'active_from']);
```
outdooracorn, line 23: That's not all this unique constraint does. It will also prevent two policies of the same type with `active_from` set to the same date. Some suggestions to make the comment clearer:

```suggestion
            // This also prevents two upcoming policies of the same type with `active_from` set to `null`
            $table->unique(['policy_type', 'active_from']);
```

or

```suggestion
            // This prevents two policies of the same type with `active_from` set to the same value (including `null` for upcoming policies)
            $table->unique(['policy_type', 'active_from']);
```

app/Policy.php
```diff
@@ -0,0 +1,52 @@
...
+    ];
+
+    protected function casts(): array {
+        return [
+            // cast `active_from` to a CarbonImmutable instance rather than a string
+            'active_from' => 'immutable_date',
+
+            'created_at' => 'immutable_datetime',
+            'updated_at' => 'immutable_datetime',
```
outdooracorn, line 50: I'm sure I tested before that `created_at` and `updated_at` returned a `Carbon` instance unless cast. Testing again now, and it seems to return a `CarbonImmutable` instance regardless of if it is cast or not. Maybe because the PR is using the default `$table->timestamps();` method, instead of manually creating the rows?

What does it return for you if you comment out these casts?

Up to you if you leave this cast in or not. Same for `PolicyAcceptance` as well.

outdooracorn (dismissed):
I've replied to your main question and left some nitpick, FYI, and suggestion comments. However, I'm generally happy with this PR and its schemas and the models.

database/migrations/2026_06_22_083853_create_policies_table.php
tarrow: Thanks for the detailed thoughts; it's quite a minefield. I also found that following a link on laravel/framework#12490 to laravel/framework#12060 was fairly informative.

For example
> but I guess this is because Eloquent correctly handles setting the value of created_at and updated_at when INSERTing or UPDATEing data.

seems to be confirmed by [a comment on that issue](https://github.com/laravel/framework/issues/12060#issuecomment-179268591).

I could have sworn I also linked this issue more explicitly somewhere but I guess I failed.

My current thinking is that the point of using eloquent as an ORM and the built in features like `timestamps()` is that we shouldn't need to think about or dig into raw SQL; at most we should only need to integration test to check they give us the behaviour we find desirable.

database/migrations/2026_06_22_083910_create_policy_acceptances_table.php
tarrow: Absolutely no reason, I simply pasted or typed it there; more than happy to move it!

tests/PolicyAcceptanceTest.php
tarrow: I'm not sure it is overkill. I think it actually exposes that the system we have made is potentially confusing.

As I read this it it making me reconsider if having `accepted_at` as guarded is the wisest, least confusing decision.

Currently if I:
- don't explicitly set `accepted_at` *then* the DB sets it to be `NOW`
- explicitly (but incorrectly) mass assign it to be a specific timestamp *then* the value is silently dropped and the DB sets it to be `NOW`
- explicitly assign it to be a specific timestamp *then* the value is stored and used

This makes me conclude that the best thing to do would probably be to make the field `fillable`. Does this make sense to you @outdooracorn? I think you introduced the initial idea of guarding it in the very first commit of this PR.

app/Policy.php
tarrow: > What does it return for you if you comment out these casts?

We should probably meet and confirm our findings; I observe (running tests locally) that without the `casts()` I get a `Carbon` for `created_at` and `updated_at`.

database/migrations/2026_06_22_083853_create_policies_table.php
tarrow: @outdooracorn and I talked and we agree we will stick with the laravel default `timestamps()`

app/Policy.php
outdooracorn: I'm an idiot!  :sweat_smile: I was testing the wrong field. :man_facepalming:

tests/PolicyAcceptanceTest.php
outdooracorn: @tarrow and I discussed this in a call and ended up deciding to:
* Use a `NON NULL DATETIME` in the database instead of using a `TIMESTAMP` that would have to be nullable.
* Allow `accepted_at` to be mass-assigned (by removing it from the `$guarded` to `$fillable` array). The mass-assignment protection doesn't protect us that much and just adds more confusion when using the model.

I've updated the code and tests to reflect this decision.

There is also an option for a follow-up that makes the model auto-set `accepted_at` if it is missing or `null`.

tests/PolicyAcceptanceTest.php
outdooracorn: This test has now been removed as we are allowing `accepted_at` to be mass-assigned now. Will resolve this thread.

tests/PolicyAcceptanceTest.php
```diff
@@ -47,14 +49,34 @@ public function testCreatesAndSavesSuccessfully(): void {
...
+            'policy_id' => $this->policyId,
+            'accepted_at' => null,
+        ]);
+    }
+
+    public function testUserAcceptingSamePolicyTwiceFails() {
+        PolicyAcceptance::create([
+            'user_id' => $this->userId,
+            'policy_id' => $this->policyId,
+            'accepted_at' => CarbonImmutable::now()->subSeconds(2),
```
dati18, line 70: ah a nice edge case
+1

dati18 (dismissed):
I think it's good to go now

outdooracorn (approved):
I pushed the last few commits, but this is good to go from my side! 🚀

dati18 (approved):
approve again

tests/PolicyAcceptanceTest.php
outdooracorn: :D

PR #1178, opened 2026-06-24
Add PolicySeeder

tarrow:
Creates a new Seeder for easily creating Policy locally. This is not automatically run with `php artisan db:seed` since I intend to follow up with separating out Seeders into those we may wish to run only in test or dev and those we may wish to run in Prod

Bug: T430090

### DIFF
--- a/database/seeds/PolicySeeder.php
+++ b/database/seeds/PolicySeeder.php
@@ -1,0 +1,19 @@
+<?php
+
+namespace Database\Seeders;
+
+use App\Policy;
+use Carbon\CarbonImmutable;
+use Illuminate\Database\Seeder;
+
+class PolicySeeder extends Seeder {
+    public function run() {
+        Policy::create(
+            [
+                'policy_type' => 'terms-of-use',
+                'active_from' => CarbonImmutable::createFromDate(2026, 06, 01),
+                'content_vue_file' => 'terms-of-use/example.vue',
+            ]
+        );
+    }
+}
--- a/app/Policy.php
+++ b/app/Policy.php
@@ -45,7 +45,6 @@ class Policy extends Model {
             // cast `active_from` to a CarbonImmutable instance rather than a string
             'active_from' => 'immutable_date',
 
-            // TODO: should we make Laravel use CarbonImmutable globally instead of casting in models?
             'created_at' => 'immutable_datetime',
             'updated_at' => 'immutable_datetime',
         ];
--- a/app/Policy.php
+++ b/app/Policy.php
@@ -8,12 +8,12 @@ use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Model;
 
 /**
- * @property-read int $id
+ * @property int $id
  * @property string $policy_type
  * @property CarbonImmutable|null $active_from
  * @property string $content_vue_file
- * @property CarbonImmutable $created_at
- * @property CarbonImmutable $updated_at
+ * @property CarbonImmutable|null $created_at
+ * @property CarbonImmutable|null $updated_at
  *
  * @method static Builder<static>|Policy newModelQuery()
  * @method static Builder<static>|Policy newQuery()
--- a/app/Policy.php
+++ b/app/Policy.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace App;
+
+use Carbon\CarbonImmutable;
+use Eloquent;
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Model;
+
+/**
+ * @property-read int $id
+ * @property string $policy_type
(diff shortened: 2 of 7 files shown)


dati18 (dismissed):

database/seeds/PolicySeeder.php
```diff
@@ -0,0 +1,19 @@
...
+
+class PolicySeeder extends Seeder {
+    public function run() {
+        Policy::create(
+            [
+                'policy_type' => 'terms-of-use',
+                'active_from' => CarbonImmutable::createFromDate(2026, 06, 01),
+                'content_vue_file' => 'terms-of-use/example.vue',
+            ]
+        );
```
outdooracorn, line 17: I believe this will create a new policy, or fail if the policy already exists. If multiple seeders are run in the future, I think a failed Seeder will prevent subsequent seeders from running, which is likely not what we want.

From my quick searching, it's often recommended to create seeders in an idempotent way (i.e. running the seeder multiple times will result in the same database state).

https://oneuptime.com/blog/post/2026-02-03-laravel-database-seeding/view recommends using `firstOrCreate()` or `updateOrCreate()` methods on the model to make seeders idempotent.

Alternatively/additionally, you could use the `Policy::truncate();` method to clear the table and reset the index (I haven't thought much about how best to handle existing constraints, though).

database/seeds/PolicySeeder.php
```diff
@@ -0,0 +1,19 @@
...
+use App\Policy;
+use Carbon\CarbonImmutable;
+use Illuminate\Database\Seeder;
+
+class PolicySeeder extends Seeder {
+    public function run() {
+        Policy::create(
+            [
+                'policy_type' => 'terms-of-use',
+                'active_from' => CarbonImmutable::createFromDate(2026, 06, 01),
```
outdooracorn, line 14: Creating a `CarbonImmutable` instance here works, but makes it trickier to use `firstOrCreate()` or `updateOrCreate()` methods (see other comment as to why you would want to use those methods). This is because (emphasis mine):

> Create a Carbon instance from just a date. **The time portion is set to now**.

And explained using tinker:

```
$ dc exec -it api php artisan tinker
> use Carbon\CarbonImmutable;

> $a = CarbonImmutable::createFromDate(2026, 06, 01);
= Carbon\CarbonImmutable @1780325575 {#8373
    date: 2026-06-01 14:52:55.291326 UTC (+00:00),
  }

> $b = CarbonImmutable::createFromDate(2026, 06, 01);
= Carbon\CarbonImmutable @1780325594 {#8399
    date: 2026-06-01 14:53:14.547747 UTC (+00:00),
  }

> $a === $b
= false

> $a == $b
= false

> $a->equalTo($b)
= false
```

To address this you could either:
* use a date string (e.g. `'active_from' => '2026-06-01',`)
* use the `CarbonImmutable::createMidnightDate()` method

```
> $a = CarbonImmutable::createMidnightDate(2026, 06, 01);
= Carbon\CarbonImmutable @1780272000 {#8474
    date: 2026-06-01 00:00:00.0 UTC (+00:00),
  }

> $b = CarbonImmutable::createMidnightDate(2026, 06, 01);
= Carbon\CarbonImmutable @1780272000 {#8451
    date: 2026-06-01 00:00:00.0 UTC (+00:00),
  }

> $a === $b
= false

> $a == $b
= true

> $a->equalTo($b)
= true
```

database/seeds/PolicySeeder.php
```diff
@@ -0,0 +1,19 @@
...
+                'content_vue_file' => 'terms-of-use/example.vue',
+            ]
+        );
+    }
```
outdooracorn, line 18: For testing purposes, would it be useful to have more than one current policy? Maybe also a superseded policy? I'm also happy if we add these at a later stage, though.

outdooracorn (changes requested):
It might be good for the PR description / commit message to include:
* how you can run this seeder (`php artisan db:seed --class=PolicySeeder`), as well as why it won't be run with `php artisan db:seed`. 
* that you might need to run `composer dump-autoload` in order for the `db:seed` command to know that the `PolicySeeder` class exists.

database/seeds/PolicySeeder.php
outdooracorn: I did some more testing of creating `CarbonImmutable` objects while writing a test for another PR and thought I would paste the findings here as well.
```
echo CarbonImmutable::create(2026, 06, 01);              // 2026-06-01 00:00:00
echo CarbonImmutable::createFromDate(2026, 06, 01);      // 2026-06-01 16:01:49
echo CarbonImmutable::createMidnightDate(2026, 06, 01);  // 2026-06-01 00:00:00
echo CarbonImmutable::create(2026, 06, 01, 12);          // 2026-06-01 12:00:00
```

PR #1182, opened 2026-07-01
Remove all terms of use logic code

rosalieper:
* Drop database tables
* Remove Laravel Jobs
* Remove any logic and tests related to Terms of Use

Bug:  T430532


tarrow (approved):
Looks good to me. Nothing to add really.

I had a second look and didn't see anything else old hanging around.

Ran the migrations locally and confirmed they clean up my tables.

I'm happy there's no `down()` for the migration. I think this makes prefect sense.

dati18 (approved):

PR #1192, opened 2026-07-08
Add initial terms of use policy via migration

AndrewKostka:
Bug: [T429449](https://phabricator.wikimedia.org/T429449)


dati18 (approved):
I checked the migration code and ran the migration on my local dev env and everything seems to be ok

PR #1197, opened 2026-07-10
Add endpoint for creating a policy acceptance for a specific user

dati18:
**Summary**
- Add authenticated endpoint to record policy acceptances for the current user, with strict validation and all-or-nothing behavior for invalid policy IDs.

**Behavior**
- Uses authenticated user from session context (no `user_id` in payload)
- Ignores already-accepted policies (idempotent writes via `firstOrCreate`)
- If any provided policy ID does not exist, returns `400` with `missing_policy_ids`
- No acceptances are created when invalid policy IDs are present
- Returns `422` for invalid payload (missing `policy_ids`, non-array, non-integer items)

Bug: T431674

### DIFF
--- a/app/Http/Controllers/PolicyAcceptanceController.php
+++ b/app/Http/Controllers/PolicyAcceptanceController.php
@@ -1,0 +1,43 @@
+<?php
+
+namespace App\Http\Controllers;
+
+use App\Policy;
+use App\PolicyAcceptance;
+use Illuminate\Http\JsonResponse;
+use Illuminate\Http\Request;
+
+class PolicyAcceptanceController extends Controller {
+    public function store(Request $request): JsonResponse {
+        $request->validate([
+            'policy_ids' => ['required', 'array'],
+            'policy_ids.*' => ['integer'],
+        ]);
+
+        $policyIds = $request->input('policy_ids');
+
+        // Check all policy IDs exist before writing anything
+        $existingIds = Policy::whereIn('id', $policyIds)->pluck('id')->all();
+        $missingIds = array_values(array_diff($policyIds, $existingIds));
+
+        if (!empty($missingIds)) {
+            return response()->json([
+                'success' => false,
+                'message' => 'Some policy IDs do not exist.',
+                'data' => ['missing_policy_ids' => $missingIds],
+            ], 400);
+        }
+
+        $userId = $request->user()->id;
+
+        foreach ($policyIds as $policyId) {
+            // Ignore if the user has already accepted this policy
+            PolicyAcceptance::firstOrCreate(
+                ['user_id' => $userId, 'policy_id' => $policyId],
+                ['accepted_at' => now()],
+            );
+        }
+
+        return response()->json(['success' => true]);
+    }
+}
--- a/routes/api.php
+++ b/routes/api.php
@@ -29,7 +29,7 @@ $router->group(['middleware' => ['throttle:45,1']], function () use ($router): v
         $router->delete('auth/login', ['uses' => 'Auth\LoginController@deleteLogin']);
 
         // policy acceptances
-        $router->put('policy_acceptances', ['uses' => 'PolicyAcceptanceController@store']);
+        $router->put('v1/policy_acceptances', ['uses' => 'PolicyAcceptanceController@store']);
 
         // user
         $router->group(['prefix' => 'user'], function () use ($router): void {
--- a/routes/api.php
+++ b/routes/api.php
@@ -28,6 +28,9 @@ $router->group(['middleware' => ['throttle:45,1']], function () use ($router): v
         $router->get('auth/login', ['uses' => 'Auth\LoginController@getLogin']);
         $router->delete('auth/login', ['uses' => 'Auth\LoginController@deleteLogin']);
 
+        // policy acceptances
+        $router->put('policy_acceptances', ['uses' => 'PolicyAcceptanceController@store']);
+
         // user
         $router->group(['prefix' => 'user'], function () use ($router): void {
             $router->post('sendVerifyEmail', ['uses' => 'UserVerificationTokenController@createAndSendForUser']);
--- a/tests/Routes/PolicyAcceptanceControllerTest.php
+++ b/tests/Routes/PolicyAcceptanceControllerTest.php
@@ -11,7 +11,7 @@ use Illuminate\Foundation\Testing\DatabaseTransactions;
 use Tests\TestCase;
 
 class PolicyAcceptanceControllerTest extends TestCase {
-    protected $route = 'policy_acceptances';
+    protected $route = 'v1/policy_acceptances';
 
     use DatabaseTransactions;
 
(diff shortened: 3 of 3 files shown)


tests/Routes/PolicyAcceptanceControllerTest.php
```diff
@@ -0,0 +1,117 @@
...
+    public function testUnauthenticatedRequestResponds401(): void {
+        $this->json('PUT', $this->route)
+            ->assertStatus(401);
+    }
+
+    public function testAcceptSinglePolicy(): void {
+        $user = User::factory()->create();
+        $policy = $this->makePolicy();
+
+        $this->actingAs($user, 'api')
```
tarrow, line 37: I had to check what `api` means in this context. Turns out it is the name of the `guard`. Is it necessary to specify the guard here?

I do see we do this in most other places (exception is LimitWikiAccessTest) so probably we should just keep this here. I don't think you should change anything but I'd be interested if you know the answer to the question.

tests/Routes/PolicyAcceptanceControllerTest.php
```diff
@@ -0,0 +1,117 @@
...
+            ->assertJsonFragment(['missing_policy_ids' => [$nonExistentId]]);
+
+        // Nothing should have been written
+        $this->assertDatabaseMissing('policy_acceptances', [
+            'user_id' => $user->id,
+            'policy_id' => $policy->id,
+        ]);
+    }
+
+    public function testMissingPolicyIdsFieldReturns422(): void {
```
tarrow, line 125: We didn't specify it in the ticket but I am happy that you kept the same behaviour as other endpoints for validation failures.

app/Http/Controllers/PolicyAcceptanceController.php
```diff
@@ -0,0 +1,43 @@
...
+        ]);
+
+        $policyIds = $request->input('policy_ids');
+
+        // Check all policy IDs exist before writing anything
+        $existingIds = Policy::whereIn('id', $policyIds)->pluck('id')->all();
+        $missingIds = array_values(array_diff($policyIds, $existingIds));
+
+        if (!empty($missingIds)) {
+            return response()->json([
```
tarrow, line 24: I looked and thought should we `return response()->json(..)` or throw here? I checked what is the standard and we regularly do this so let's keep this pattern.

tests/Routes/PolicyAcceptanceControllerTest.php
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    public function testPolicyIdsNotAnArrayReturns422(): void {
+        $user = User::factory()->create();
+
+        $this->actingAs($user, 'api')
+            ->json('PUT', $this->route, ['policy_ids' => 'not-an-array'])
+            ->assertStatus(422)
+            ->assertJsonStructure(['errors' => ['policy_ids']]);
+    }
```
tarrow, line 141: I think these tests are very thorough but I did also realise you could test for the "non-integer policy id" situation that you mentioned in the PR description.

Not a blocker to merging if you don't want to add it though.

app/Http/Controllers/PolicyAcceptanceController.php
```diff
@@ -0,0 +1,43 @@
...
+            ], 400);
+        }
+
+        $userId = $request->user()->id;
+
+        foreach ($policyIds as $policyId) {
+            // Ignore if the user has already accepted this policy
+            PolicyAcceptance::firstOrCreate(
+                ['user_id' => $userId, 'policy_id' => $policyId],
+                ['accepted_at' => now()],
```
tarrow, line 37: Does this do what we want? By reading it (but not executing and trying it out) I'm worried this updates the `accepted_at` when it should remain at the old time. Perhaps we could confirm this by updating the test (comment added below)

tests/Routes/PolicyAcceptanceControllerTest.php
```diff
@@ -0,0 +1,117 @@
...
+            'policy_id' => $policy->id,
+            'accepted_at' => now(),
+        ]);
+
+        $this->actingAs($user, 'api')
+            ->json('PUT', $this->route, ['policy_ids' => [$policy->id]])
+            ->assertStatus(200)
+            ->assertJson(['success' => true]);
+
+        $this->assertSame(1, PolicyAcceptance::where([
```
tarrow, line 77: Could we test that the `accepted_at` value isn't changed here? (see above)

tarrow (changes requested):
Only blocker from my side is around confirming we don't update `accepted_at` time on each request.

tests/Routes/PolicyAcceptanceControllerTest.php
dati18: The second argument `api` is telling Laravel's auth manager that "for this test duration, set this user as an _authenticated user_ for the `api` guard". When you just call `$this->actingAs($user)`, Laravel logs that user into default guard (I haven't checked what is the "default", usually cookie auth?).
We have been using this pattern for testing purposes, basically bypassing all the hassle of authenticating the normal way

tests/Routes/PolicyAcceptanceControllerTest.php
dati18: As I understand, `firstOrCreate()` function doesn't do row update if row is already there. Check [this](https://laravel.com/docs/13.x/eloquent#:~:text=or%20Creating%20Models-,The,merging%20the%20first%20array%20argument%20with%20the%20optional%20second%20array%20argument.,-The)

But I kind of understood your standpoint, so I created a test for edge case like this.

tests/Routes/PolicyAcceptanceControllerTest.php
tarrow: Yes, I was wondering if we could either use the default guard or make `api` the default guard. You're right this is the same pattern we use in other places and is a good pattern to follow I think

tests/Routes/PolicyAcceptanceControllerTest.php
tarrow: Thanks, looks like ` updateOrCreate()` would have been the function that would get and potentially update it.

routes/api.php
```diff
@@ -28,6 +28,9 @@
         $router->get('auth/login', ['uses' => 'Auth\LoginController@getLogin']);
         $router->delete('auth/login', ['uses' => 'Auth\LoginController@deleteLogin']);
 
+        // policy acceptances
+        $router->put('policy_acceptances', ['uses' => 'PolicyAcceptanceController@store']);
```
tarrow: I realise I missed one small thing in my previous review (sorry). The ticket asks for this api to be under `/v1/` rather than in the root.

This is a pattern we are not yet following everywhere but something we would like to do.
```suggestion
        $router->put('v1/policy_acceptances', ['uses' => 'PolicyAcceptanceController@store']);
```

dati18: ah a blunder... fixed

tarrow (approved):
Thanks for the final change. I'm now happy with this.

PR #1198, opened 2026-07-10
Create endpoint to get polices that should be accepted

rosalieper:
Bug: T429591


app/Http/Controllers/PoliciesController.php
```diff
@@ -0,0 +1,23 @@
...
+use App\Http\Resources\PoliciesCollection;
+use App\Policy;
+use Carbon\CarbonImmutable;
+
+class PoliciesController extends Controller {
+    public function getCurrentPolicies() {
+        $now = CarbonImmutable::now();
+
+        // This works based on the assumption that the latest policy has the highest id given that id is AUTO_INCREMENT
+        $latestPolicyIds = Policy::where('active_from', '<', $now)
```
dati18, line 14: As I understand, query 1 builds a list of IDs with group-by + max and then `pluck`, query 2 fetches rows by those IDs.
So it means you do 2 round trip to get IDs and fetch rows. I think for a small set  of data it's fine, but I recommend avoiding this pattern for large data set

app/Http/Controllers/PoliciesController.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace App\Http\Controllers;
+
+use App\Http\Resources\PoliciesCollection;
+use App\Policy;
+use Carbon\CarbonImmutable;
+
+class PoliciesController extends Controller {
+    public function getCurrentPolicies() {
```
dati18: Consider adding a `return` type (aka `PoliciesCollection` in this case)

app/Http/Controllers/PoliciesController.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace App\Http\Controllers;
+
+use App\Http\Resources\PoliciesCollection;
+use App\Policy;
+use Carbon\CarbonImmutable;
+
```
dati18, line 8: consider adding PHPDoc for this controller

app/Http/Resources/PoliciesCollection.php
```diff
@@ -0,0 +1,19 @@
...
+
+class PoliciesCollection extends ResourceCollection {
+    /**
+     * Transform the resource collection into an array.
+     *
+     * @return array<int|string, mixed>
+     */
+    public function toArray(Request $request): array {
+        return [
+            'items' => $this->collection,
```
dati18: the key `items` sounds "off" for me. `ResourceCollection`'s responses look like `{ "data": [ // stuff go here] }`.

Not a functionality issue but coding style

https://laravel.com/docs/12.x/collections

tests/Http/Controllers/PoliciesControllerTest.php
```diff
@@ -0,0 +1,27 @@
...
+        ]);
+        // Active policy
+        Policy::factory()->create([
+            'active_from' => now()->subMonth(),
+        ]);
+
+        $response = $this->getJson('/policies/current');
+
+        $response->assertOk();
+        $response->assertJsonCount(1, 'data.items');
```
dati18: this test only asserts the count, so it can pass even if the endpoint returns the wrong policy. 
You should add assertions that the returned item is the expected 'active' policy (by `id` or `active_from` value) and that the "future" policies are left out.

tests/Http/Controllers/PoliciesControllerTest.php
```diff
@@ -0,0 +1,27 @@
...
+use Illuminate\Foundation\Testing\DatabaseTransactions;
+use Tests\TestCase;
+
+class PoliciesControllerTest extends TestCase {
+    use DatabaseTransactions;
+
+    public function testGetCurrentPolicies(): void {
+        // Future policy
+        Policy::factory()->create([
+            'active_from' => now()->addDay(),
```
dati18: real-time `now()` calls can be fragile when used on different timestamps.
Freeze time with `Carbon::setTestNow()` or just a plain simple `$currentTime = now()`

tests/Http/Controllers/PoliciesControllerTest.php
dati18, line 1: a few suggestions for extra test coverage:
1. Returns one current policy per policy type
  - Create two terms-of-use rows and two hosting-policy rows, with mixed active dates, then assert exactly 2 results (one per type).
2. Picks latest active policy for each type
  - For the same type, create old active, newer active and future active
  - Assert the returned row is the newer active, not old and not future.
3. Returns one current policy per policy type
  - Create two `terms-of-use` rows and two `hosting-policy` rows, with mixed active dates, then assert exactly 2 results (one per type).
4. Current query uses MAX(id) per type (in `PoliciesController.php`). Add a test that backfills an older `active_from` with higher `id` to document expected behavior.

app/Http/Resources/PoliciesCollection.php
rosalieper: Here it is following the example from the ticket. but we can also do `return [$this->collection]` it will yield the output you are suggesting.

tests/Http/Controllers/PoliciesControllerTest.php
rosalieper: In this case it is not an issue but I understand what you mean.

tests/Http/Controllers/PoliciesControllerTest.php
rosalieper: yeah, that's a good point

database/factories/PolicyFactory.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace Database\Factories;
+
+use App\Policy;
+use Illuminate\Database\Eloquent\Factories\Factory;
+use Illuminate\Database\Eloquent\Model;
+
+/**
+ * @extends Factory<Model>
```
dati18: Nitpicking:

```suggestion
 * @extends Factory<Policy>
```

tests/Http/Controllers/PoliciesControllerTest.php
```diff
@@ -0,0 +1,47 @@
...
+            'policy_type' => 'terms-of-use',
+            'active_from' => $currentTime->addDay(),
+        ]);
+        // Old policy
+        Policy::factory()->create([
+            'policy_type' => 'hosting-policy',
+            'active_from' => $currentTime->subMonth(),
+        ]);
+        // Active policies
+        $latestToUPolicy = Policy::factory()->create([
```
dati18: Test strengthening idea:

The variable name `$latestToUPolicy` suggests the endpoint should return the latest active policy per type, but the test only creates one active `terms-of-use` policy. Consider adding another active terms-of-use policy with a more recent active_from to verify the selection logic actually returns the latest, not just any active policy.
And change the var name to `$latestActiveToUPolicy`

rosalieper: There can only be one valid active Terms of Use version at a time, any other version is not a valid active version. the test creates 2 versions and checks that the returned one is the right one. 

I wanted to test that it returns the right version in a case where we have records with future active_from and older active_from dates. So i did 2 HP version in the past and expect it to return the latest as the active one.

For the Term of Use I made a future and past version case and it is expected to return the latest active version not in the future.

with this now we can be sure it covers Old and Future active_from cases.

app/Http/Resources/PoliciesCollection.php
deer-wmde: I agree with Dat and would challenge the proposed example structure from the ticket. The Collection is already wrapped in a `data` object so currently we have `.data.items` which I think is redundant, and `data` gets used already in other parts of the API

rosalieper: I made the change :)

dati18 (approved):
LGTM now.

PR #1203, opened 2026-07-14
Add migration to backfill terms of use

AndrewKostka:
Bug: [T429452](https://phabricator.wikimedia.org/T429452)

### DIFF
--- a/database/migrations/2026_07_13_075347_backfill_policy_acceptance_existing_users.php
+++ b/database/migrations/2026_07_13_075347_backfill_policy_acceptance_existing_users.php
@@ -19,25 +19,23 @@ return new class() extends Migration {
         $timestamp = now();
 
         DB::table('users')
-            ->leftJoin('policy_acceptances', fn ($join) =>
-                $join->on('users.id', '=', 'user_id')
-                    ->where('policy_id', '=', $policyId)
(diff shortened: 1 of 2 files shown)


database/migrations/2026_07_13_075347_backfill_policy_acceptance_existing_users.php
```diff
@@ -0,0 +1,61 @@
...
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Support\Facades\DB;
+
+return new class() extends Migration {
+    /**
+     * Users created before this timestamp are treated as having accepted
+     * our existing terms of use. Users created afterwards go through our
+     * new policy flow (T430529) and are excluded from the backfill.
+     */
+    private const USER_CREATED_AT_CUTOFF = '2026-07-13 08:00:00';
```
AndrewKostka: **Do Not Merge** until [T430529](https://phabricator.wikimedia.org/T430529) is deployed on production, which will give us an accurate cutoff timestamp.

dati18 (dismissed):
I pre-approved so it can be deployed after https://phabricator.wikimedia.org/T430529 is done without any deplay

tarrow (approved):
approving as per above comments, didn't check a second time

PR #1207, opened 2026-07-17
Added controller method that returns not accepted policies by user

dati18:
**Summary**
- Controller logic updated in `PoliciesController.php` with new endpoint `getMissingPolicies`
- API route added in `routes/api.php` (new authenticated route `GET v1/policies/missing`, handled by `PoliciesController@getMissingPolicies`)
- New test class covering both current and missing policies, including coverage for:
  - auth required on missing endpoint
  - only latest active policy per type is returned
  - future policies are excluded
  - already accepted policies are excluded for current user
  - another user’s acceptances do not affect current user
  - empty list when all current policies are accepted
  - current policies response structure and payload fragments
- The older `PoliciesController` test under `tests/Http/Controllers` was merged into the route-level file above because they do full http behaviors like auth middleware, URL, JSON request/response, calling endpoints, etc., not isolated controller functions

Bug: T432337


app/Http/Controllers/PoliciesController.php
```diff
@@ -20,4 +22,23 @@ public function getCurrentPolicies(): PoliciesCollection {
...
+
+    public function getMissingPolicies(Request $request): PoliciesCollection {
+        $now = CarbonImmutable::now();
+
+        $activePolicyIds = Policy::where('active_from', '<=', $now)
+            ->selectRaw('MAX(id) as id')
+            ->groupBy('policy_type')
+            ->pluck('id');
+
+        $acceptedPolicyIds = PolicyAcceptance::where('user_id', $request->user()->id)
```
rosalieper: what do you think of renaming this to `$userAcceptedPolicyIds` just for clarity.

app/Http/Controllers/PoliciesController.php
```diff
@@ -20,4 +22,23 @@ public function getCurrentPolicies(): PoliciesCollection {
 
         return new PoliciesCollection($currentPolicies);
     }
+
+    public function getMissingPolicies(Request $request): PoliciesCollection {
+        $now = CarbonImmutable::now();
+
+        $activePolicyIds = Policy::where('active_from', '<=', $now)
```
rosalieper: since we have this block of code twice here, I think it could be nice to have a private function for this query and call the it both in `getCurrentPolicies` and `getMissingPolicies`.

tests/Routes/PoliciesControllerTest.php
rosalieper, line 1: there are now two `polliciesControllerTest.php`. one in the `controller` tests dir and this one. Will be good to merge this two test files because they contain tests from the same class.

dati18: merged and moved to `/test/Routes` because they do full http behaviors like auth middleware, URL, JSON request/response, calling endpoints, etc., not isolated controller functions

rosalieper (dismissed):
~I think you have accidentally added unwanted changes to this PR. I see changes from  #1207 here.~  ignore

rosalieper (approved):
I am happy with the change 💯

app/Http/Controllers/PoliciesController.php
```diff
@@ -5,19 +5,37 @@
...
-        $currentPolicies = Policy::whereIn('id', $latestPolicyIds)->get();
+        $currentPolicies = Policy::whereIn('id', $this->activePolicyIdsQuery($now))->get();
 
         return new PoliciesCollection($currentPolicies);
     }
+
+    public function getMissingPolicies(Request $request): PoliciesCollection {
+        $now = CarbonImmutable::now();
+
+        $missingPolicies = Policy::whereIn('id', $this->activePolicyIdsQuery($now))
```
tarrow, line 23: I spent a while looking at this to try and see if we could find a way to not some some raw query and try to find something more readable. One vague direction I came up with was something like this:

```suggestion
        $missingPolicies = Policy::whereIn('id', $this->activePolicyIdsQuery($now))
        ->whereDoesntHave('acceptances', function (EloquentBuilder $query) use ($request) {
            $query->where('user_id', $request->user()->id);
        })->get();
```
by also adding an acceptances relation on to the policy like this:
```
public function acceptances(): HasMany {
        return $this->HasMany(PolicyAcceptance::class);
    }
```
I think this is probably a more ORM focused way to approach the issue. However given we also use raw sql in the query to get the activePolicyIds I don't think being inconsistent here buys us much.

tarrow (approved):
I really like a lot of this patch.

The tests are super comprehensive and readable to me.

I liked that you identified that this should be in `Routes` test.

I'm still a bit sceptical about the query logic which I personally find hard a little harder to parse than the far less efficient first draft but I do agree that performance probably matters here so it's an acceptable tradeoff for me.

I've left a small suggestion for the direction I looked into *just* for your interest but I don't think we should look further into this.

PR #1210, opened 2026-07-20
Create endpoint to get specific policy by type and active_from date

rosalieper:
Bug: T432339

### DIFF
--- a/app/Http/Controllers/PoliciesController.php
+++ b/app/Http/Controllers/PoliciesController.php
@@ -3,11 +3,8 @@
 namespace App\Http\Controllers;
 
 use App\Http\Resources\PoliciesCollection;
-use App\Http\Resources\PolicyResource;
 use App\Policy;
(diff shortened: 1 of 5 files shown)


tarrow (approved):
looks fine to me; paired with @rosalieper on this; took a short look away and then back again.

I was happy that we reused the Resource; this feels like exactly what we intended to be possible to do.

I was also happy with our decisions to use the Validator and its `safe()` function.

I'm slightly less happy with the global `abort()`. I feel like it would have been nicer to call this on an injected Request but this does follow the very terse advice on [Error Handling](https://laravel.com/docs/11.x/errors#custom-http-error-pages).

We also read about [Resource Controllers](https://laravel.com/docs/11.x/controllers#resource-controllers) and wondered if this is a pattern we ought to adopt in the future. For now we decided not to look at this further and inject another interesting but inconsistent pattern.

PR #1215, opened 2026-07-24
fix PolicyResource: `null` date should not default to valid date

deer-wmde:
fix for the discovery in https://github.com/wmde/wbaas-deploy/pull/2925

https://phabricator.wikimedia.org/T432714

### DIFF
--- a/app/Http/Resources/PolicyResource.php
+++ b/app/Http/Resources/PolicyResource.php
@@ -15,11 +15,17 @@ class PolicyResource extends JsonResource {
     public static $wrap = null;
 
     public function toArray(Request $request): array {
+        $activeFrom = $this->active_from;
+
+        if ($this->active_from !== null) {
+            $activeFrom = Carbon::parse($this->active_from)->format('Y-m-d');
+        }
+    
         return [
             'metadata' => [
                 'policy_id' => $this->id,
                 'type' => $this->policy_type,
-                'active_from' => Carbon::parse($this->active_from)->format('Y-m-d'),
+                'active_from' => $activeFrom,
                 'content_vue_file' => $this->content_vue_file,
             ],
         ];
--- a/app/Http/Resources/PolicyResource.php
+++ b/app/Http/Resources/PolicyResource.php
@@ -20,7 +20,7 @@ class PolicyResource extends JsonResource {
         if ($this->active_from !== null) {
             $activeFrom = Carbon::parse($this->active_from)->format('Y-m-d');
         }
-    
+
         return [
             'metadata' => [
                 'policy_id' => $this->id,
--- a/app/Http/Resources/PolicyResource.php
+++ b/app/Http/Resources/PolicyResource.php
@@ -25,7 +25,9 @@ class PolicyResource extends JsonResource {
             'metadata' => [
                 'policy_id' => $this->id,
                 'type' => $this->policy_type,
-                'active_from' => $activeFrom,
+                'active_from' => $this->active_from === null
+                    ? null
+                    : Carbon::parse($this->active_from)->format('Y-m-d'),
                 'content_vue_file' => $this->content_vue_file,
             ],
         ];
--- a/app/Http/Resources/PolicyResource.php
+++ b/app/Http/Resources/PolicyResource.php
@@ -15,12 +15,6 @@ class PolicyResource extends JsonResource {
     public static $wrap = null;
 
     public function toArray(Request $request): array {
-        $activeFrom = $this->active_from;
-
-        if ($this->active_from !== null) {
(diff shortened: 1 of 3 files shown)


app/Http/Resources/PolicyResource.php
```diff
@@ -15,11 +15,17 @@ class PolicyResource extends JsonResource {
...
+        if ($this->active_from !== null) {
+            $activeFrom = Carbon::parse($this->active_from)->format('Y-m-d');
+        }
+
         return [
             'metadata' => [
                 'policy_id' => $this->id,
                 'type' => $this->policy_type,
-                'active_from' => Carbon::parse($this->active_from)->format('Y-m-d'),
+                'active_from' => $activeFrom,
```
dati18: I think we can write it on the same line like this. It's your choice so you don't have to take the suggestion :D

```suggestion
                'active_from' => $this->active_from === null
                    ? null
                    : Carbon::parse($this->active_from)->format('Y-m-d'),
```

tests/Resources/PolicyResourceTest.php
```diff
@@ -0,0 +1,43 @@
...
+            'policy_type' => 'terms-of-use',
+            'content_vue_file' => 'terms-of-use/example.vue',
+            'active_from' => '2022-02-02',
+        ]);
+
+        $resource = new PolicyResource($policy);
+        $data = json_decode($resource->toJson());
+
+        $this->assertEquals(
+            data_get($data, 'metadata.active_from'),
```
dati18: I think you are passing parameters in the wrong order, it should be: `assertEquals(expected_value, actual_value)`
Swap the arguments so the expected value is first and `data_get()` is second

tests/Resources/PolicyResourceTest.php
```diff
@@ -0,0 +1,43 @@
...
+        $policy = Policy::create([
+            'policy_type' => 'terms-of-use',
+            'content_vue_file' => 'terms-of-use/example.vue',
+        ]);
+
+        $resource = new PolicyResource($policy);
+        $data = json_decode($resource->toJson());
+
+        $this->assertEquals(
+            data_get($data, 'metadata.active_from'),
```
dati18, line 40: same issue as line 24

tests/Resources/PolicyResourceTest.php
dati18: If you want to assert the exact type/representation produced by the resource (especially for dates), consider using `assertSame`

tests/Resources/PolicyResourceTest.php
```diff
@@ -0,0 +1,43 @@
...
+
+    public function testActiveFrom(): void {
+        $policy = Policy::create([
+            'policy_type' => 'terms-of-use',
+            'content_vue_file' => 'terms-of-use/example.vue',
+            'active_from' => '2022-02-02',
+        ]);
+
+        $resource = new PolicyResource($policy);
+        $data = json_decode($resource->toJson());
```
dati18: Testing the resource via `toJson() + json_decode()` seems clunky for me. I prefer asserting against the array form of the resource.

tests/Resources/PolicyResourceTest.php
dati18, line 1: I'd like to see a test that verify `metadata.active_from` is null for upcoming policy with `active_from = null`, which make the test stronger and prevent mismatches

tests/Resources/PolicyResourceTest.php
deer-wmde: ah good catch!

tests/Resources/PolicyResourceTest.php
deer-wmde: true; i wanted to make sure whatever transformations happen, happen. Looking a bit into the underlying class led me to the `resolve` method that we can use here https://api.laravel.com/docs/11.x/Illuminate/Http/Resources/Json/JsonResource.html#method_resolve

app/Http/Resources/PolicyResource.php
```diff
@@ -15,11 +15,19 @@ class PolicyResource extends JsonResource {
     public static $wrap = null;
 
     public function toArray(Request $request): array {
+        $activeFrom = $this->active_from;
```
dati18: remove this block then I think it's good to go

tests/Resources/PolicyResourceTest.php
deer-wmde: good call, i added a test case where its set explicitly to null. I think its good to keep the other one so that we can be assured this is the behavior even if omitted

dati18 (approved):
awesome work and collab. Approved!

PR #1221, opened 2026-07-30
fix(api): make PollForMediaWikiJobsJob skip soft-deleted wikis

dati18:
Bug: T433575

### DIFF
--- a/app/Jobs/PollForMediaWikiJobsJob.php
+++ b/app/Jobs/PollForMediaWikiJobsJob.php
@@ -24,7 +24,7 @@ class PollForMediaWikiJobsJob extends Job implements ShouldBeUnique, ShouldQueue
         foreach ($wikis as $wiki) {
             try {
                 $backendUrl = $this->mwHostResolver->getBackendUrlForWiki($wiki);
-            } catch (UnknownWikiDomainException | UnknownDBVersionException $e) {
+            } catch (UnknownWikiDomainException|UnknownDBVersionException $e) {
                 Log::warning('Skipping wiki ' . $wiki->domain . ' for pending MediaWiki jobs: ' . $e->getMessage());
 
                 continue;
--- a/app/Jobs/PollForMediaWikiJobsJob.php
+++ b/app/Jobs/PollForMediaWikiJobsJob.php
@@ -3,6 +3,7 @@
 namespace App\Jobs;
 
 use App\Services\MediaWikiHostResolver;
+use App\Services\UnknownWikiDomainException;
 use App\Wiki;
 use Illuminate\Contracts\Queue\ShouldBeUnique;
 use Illuminate\Contracts\Queue\ShouldQueue;
@@ -16,7 +17,7 @@ class PollForMediaWikiJobsJob extends Job implements ShouldBeUnique, ShouldQueue
 
     public function handle(MediaWikiHostResolver $mwHostResolver): void {
         $this->mwHostResolver = $mwHostResolver;
-        $allWikiDomains = Wiki::all()->pluck('domain');
+        $allWikiDomains = Wiki::whereNull('deleted_at')->pluck('domain');
         foreach ($allWikiDomains as $wikiDomain) {
             if ($this->hasPendingJobs($wikiDomain)) {
                 $this->enqueueWiki($wikiDomain);
@@ -25,11 +26,17 @@ class PollForMediaWikiJobsJob extends Job implements ShouldBeUnique, ShouldQueue
     }
 
     private function hasPendingJobs(string $wikiDomain): bool {
-        $response = Http::withHeaders([
-            'host' => $wikiDomain,
-        ])->get(
-            $this->mwHostResolver->getBackendUrlForDomain($wikiDomain) . '/w/api.php?action=query&meta=siteinfo&siprop=statistics&format=json'
-        );
+        try {
+            $response = Http::withHeaders([
+                'host' => $wikiDomain,
+            ])->get(
+                $this->mwHostResolver->getBackendUrlForDomain($wikiDomain) . '/w/api.php?action=query&meta=siteinfo&siprop=statistics&format=json'
+            );
+        } catch (UnknownWikiDomainException $e) {
+            Log::warning('Skipping wiki ' . $wikiDomain . ' for pending MediaWiki jobs: ' . $e->getMessage());
+
+            return false;
+        }
 
         if ($response->failed()) {
             $this->job->markAsFailed();
--- a/app/Jobs/PollForMediaWikiJobsJob.php
+++ b/app/Jobs/PollForMediaWikiJobsJob.php
@@ -3,6 +3,7 @@
 namespace App\Jobs;
 
 use App\Services\MediaWikiHostResolver;
+use App\Services\UnknownDBVersionException;
 use App\Services\UnknownWikiDomainException;
 use App\Wiki;
 use Illuminate\Contracts\Queue\ShouldBeUnique;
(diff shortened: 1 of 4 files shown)


app/Jobs/PollForMediaWikiJobsJob.php
```diff
@@ -16,7 +17,7 @@ class PollForMediaWikiJobsJob extends Job implements ShouldBeUnique, ShouldQueue
 
     public function handle(MediaWikiHostResolver $mwHostResolver): void {
         $this->mwHostResolver = $mwHostResolver;
-        $allWikiDomains = Wiki::all()->pluck('domain');
+        $allWikiDomains = Wiki::whereNull('deleted_at')->pluck('domain');
```
tarrow: doesn't `::all()` already exclude deleted Wikis?

app/Jobs/PollForMediaWikiJobsJob.php
```diff
@@ -25,11 +26,17 @@ public function handle(MediaWikiHostResolver $mwHostResolver): void {
...
-            $this->mwHostResolver->getBackendUrlForDomain($wikiDomain) . '/w/api.php?action=query&meta=siteinfo&siprop=statistics&format=json'
-        );
+        try {
+            $response = Http::withHeaders([
+                'host' => $wikiDomain,
+            ])->get(
+                $this->mwHostResolver->getBackendUrlForDomain($wikiDomain) . '/w/api.php?action=query&meta=siteinfo&siprop=statistics&format=json'
+            );
+        } catch (UnknownWikiDomainException $e) {
+            Log::warning('Skipping wiki ' . $wikiDomain . ' for pending MediaWiki jobs: ' . $e->getMessage());
```
tarrow: Does this help us much? Are we just turning an error into a warning? Maybe this is still useful though.

I guess the question is what causes this to happen? Such a long runtime for this job that race conditions are common?

dati18: Good question.

It does turn one failure mode from error to warning, but importantly it prevents a single stale domain from sabotaging the whole polling run.

To answer "Why this happens":
1. The job reads a list of domains.
2. Later, for each domain, resolver does another lookup.
3. If a wiki is soft-deleted (or otherwise missing) between those moments, resolver throws `UnknownWikiDomainException`.

The root cause I think is "double-read on mutable data" under multiple updates. The longer the runtime, the longer the gap between 1 and 3, the higher the probability. So basically it's the gap between the time of "reading the data" vs "using the data"

app/Jobs/PollForMediaWikiJobsJob.php
dati18: Yeah you're right... it's kinda unnecessary

app/Jobs/PollForMediaWikiJobsJob.php
tarrow: nice! Sounds like we are getting closer to the root cause.

So could we refactor the job so we do this look up earlier? Or even in a single query?

I guess there might still be a case where the wiki really has now gone away

dati18: Good idea, so you want to weed out bad domains instead of just "skipping" it?

tarrow: I think you are correct that there is a double read. Actually there is a triple read:
1. once to get the list of Wikis
1. once to get the backendHost
1. once when then MW backendHost determines the configuration of the Wiki

I am suggesting we merge the first two cases into one and get the correct backend hosts at the same time as we get the list of Wikis. There is still a chance that by the time we get to polling the Wikis that the 3rd read fails because it has been deleted but this is still a better solution.

Are you sure that marking the job as failed is sabotaging the whole run? I assumed that the job would still iterate over all Wikis but it would retry (sooner than planned) because it was marked as failed.

dati18: I dug a bit deeper, and I agree on 2 points:
- "triple read" is actually correct
- merging reads 1 and 2 is a good idea (and better design) than simply catching exception

The only point I'm not sure that I can agree with is that [markedAsFailed()](https://github.com/wbstack/api/blob/0d9873ca13cc3d41b44675f27c293afc83d3498e/app/Jobs/PollForMediaWikiJobsJob.php#L35) in `PollForMediaWikiJobsJob.php`  doesn't abort the current `foreach` loop and trigger a retry automatically by itself. There is no exception thrown, so the job handler completes, the jobs queue removes the job from the queue, and it ends. So if I understand it correctly, `markedAsFailed` here is basically a status flag, not a loop controller

So I refactored the job to resolve each wiki's backend URL once up front, so it doesn't "re-query" the resolver and the database per wiki during the poll loop

`PollForMediaWikiJobsJob.php`:
- The job now loads the active wikis once, using `Wiki::with('wikiDb')->get()`
- Instead of resolving the backend URL inside each polling step, it resolves it once per wiki while iterating

`MediaWikiHostResolver.php`:
- Added new helpers: 
  - `getBackendUrlForWiki()`, so the job can pass in a loaded wiki model directly
  - `getBackendHostForWiki()`, to compute the backend host from the wiki object
  
-> The job now does one read of wiki data and one host-resolving per wiki, rather than re-querying resolver state again and again
-> That reduces the chance of hitting stale state between the initial wiki list and the later backend host resolution

app/Services/MediaWikiHostResolver.php
```diff
@@ -44,19 +44,36 @@ public function getBackendUrlForDomain(string $domain): string {
         return 'http://' . $this->getBackendHostForDomain($domain);
     }
 
+    public function getBackendUrlForWiki($wiki): string {
+        return 'http://' . $this->getBackendHostForWiki($wiki);
+    }
+
+    public function getBackendHostForWiki($wiki): string {
+        // TODO: Make host format configurable for flexibility
+        return sprintf('mediawiki-%s-app-backend.default.svc.cluster.local', $this->getMwVersionForWiki($wiki));
```
tarrow: I agree with the TODO that this should eventually be configurable which I assume was copied from above

I don't think we want to duplicate this mapping though; can you see a way to keep just one map?

tarrow (commented):
looks close to ready to me apart from the new duplication in the resolver; I wonder if we need to any more tests or do you think that's not necessary?

app/Services/MediaWikiHostResolver.php
dati18: I did the following:
- Combine hostname generation in one helper function: `getHostForService($mwVersion, $service)`
- all callers reuse that helper instead of repeating the same `sprintf(...)` logic
-> removes the repeated service-host formatting. Behavior stays the same

tarrow:
Thanks for removing the duplication

> I wonder if we need to any more tests or do you think that's not necessary?

Did you have any thoughts about this?

dati18:
@tarrow Sorry I couldn't reply sooner because of the outage yesterday. My 2cent is that the extra test would be testing `getHostForDomain()` returning all expected service hosts using the same helper function, so it's not about testing a behavior change, but a formatting test.
All existing focused tests cover the real contract of `MediaWikiHostResolver.php`, for example, successful backend host resolution or backend URL generation, etc. That is enough because the change was only about centralizing the host-formatting logic, not changing the mapping behavior itself.

tarrow:
We talked about this in the daily @dati18 thinks that no more testing is required as per the above comment.

I'm still somewhat hesitant to approve this without doing some more thinking / manual testing to improve my confidence; I don't want us to get distracted from our main work if this Job running blows up (although by eye I don't see why it should)

Specifically I think it might be good to have tests that:
- cover any new public functions added to `MediaWikiHostResolver`
- cover the functionality of `PollForMediaWikiJobsJob` when errors are thrown by `MediaWikiHostResolver`

I didn't yet manage to find the time for some manual testing or to write these tests myself to improve my confidence.

dati18:
@tarrow I took your hints and added a few test cases based on existing tests.

**MediaWikiHostResolverTest:**
- `testResolverReturnsAllHostsForDomain` — verifies `getHostsForDomain()` returns all service hosts with correct names
 `testResolverBuildsBackendUrlForWiki` — verifies `getBackendUrlForWiki()` builds the correct `http://` URL when given a Wiki object.
 
 
 **PollForMediaWikiJobsJobTest:**
- `testSkipsWikiWhenResolverThrowsAndContinuesProcessing` — verifies that when `MediaWikiHostResolver `throws `UnknownDBVersionException` for one wiki, the job logs a warning and skips it, without failing the whole job, then continues dispatching `ProcessMediaWikiJobsJob` for the remaining healthy wiki(s).

tarrow (approved):
looks good to me; let's make sure when we ship this to staging we ensure we properly try it though.

Make sure there are some jobs created on Wikis and that they run before we promote it to production.

tarrow (approved):

